PR #131, opened 2014-12-08
do not swallow exceptions for disk checks

cburroughs:
`WriteableDiskCheck` can fail for a variety of reasons besides the
underlying `open` or `utimensat` syscalls failing.  For example:
another directory (such as `/bin` or `/lib`) can not be read, a
shared library could be missing, a ulimit reached, or a bug in the
check itself.  To allow those situations to be debugged we need to
exception.

NOTE: Log is at WARN to preserve 'at least one path must be writeable'
semantics.  However, those semantics are nuts and should probably be
changed to 'all paths must be writeable'.

### DIFF
--- a/hydra-data/src/main/java/com/addthis/hydra/util/WriteableDiskCheck.java
+++ b/hydra-data/src/main/java/com/addthis/hydra/util/WriteableDiskCheck.java
@@ -18,7 +18,12 @@ import java.io.File;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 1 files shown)


mspiegel:
Can we revisit this? It looks like this pull request is better than the current behavior which is to drop exceptions on the floor.

tea-dragon:
- we aren't entirely sure this catches any (not to mention all) cases we care about, and if it doesn't, then it is definitely misleading to future maintainers
- in the use case where it does the, admittedly terrible, thing of passing if any one check succeeds, then this can cause an arbitrarily high amount of log spam that is presumably completely uninteresting

mspiegel:
This has been superseded by https://github.com/addthis/hydra/pull/151

PR #140, opened 2015-01-09
Filter initialization is now explicit. Ref T51853.

mspiegel:
Bundle filters and value filters have a lazy initialization
mechanism that is poorly implemented and a bad design.
A developer writing a new filter must remember to
explicitly call the initialization method and it is
also possible to invoke other filters and bypass
the initialization.

In this implementation all filters have an open()
method. The application is responsible for ensuring
that open() is called before using the filter.

In this implementation the majority of initialization
is be performed in the constructor. Any initialization
that is too expensive to perform when the object
is deserialized (such as accessing the network or
the filesystem) should be placed in the open() method.
The use of SuperCodable for filters is discouraged.

### DIFF
--- a/hydra-filters/src/main/java/com/addthis/hydra/data/filter/bundle/BundleFilterFirstValue.java
+++ b/hydra-filters/src/main/java/com/addthis/hydra/data/filter/bundle/BundleFilterFirstValue.java
@@ -47,7 +47,7 @@ public class BundleFilterFirstValue extends BundleFilter {
      * An array of bundle field names to search. This field is required.
      */
     @FieldConfig(codable = true, required = true)
-    private AutoField[] in;
+    private CachingField[] in;
 
     /**
      * Output destination field. This field is required.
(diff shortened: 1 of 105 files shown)


hydra-filters/src/main/java/com/addthis/hydra/data/filter/bundle/BundleFilterFirstValue.java
```diff
@@ -104,10 +77,10 @@ public boolean filterExec(Bundle bundle) {
                     ValueArray arr = v.asArray();
                     if (arr.size() > 0) {
                         if (which != null) {
-                            ValueObject fieldName = ValueFactory.create(bound[i].getName());
-                            bundle.setValue(bound[end + 1], fieldName);
+                            ValueObject fieldName = ValueFactory.create(((CachingField) in[i]).name);
```
tea-dragon: you can probably declare the field directly as a `CachingField` if that is going to be enforced later anyway

tea-dragon:
about a third of the way through, and it looks great so far.

tea-dragon:
looks good to me

PR #141, opened 2015-01-21
Allow PathQuery#childMatch to return map of values.

mspiegel:
Replace the PathQuery#childMatch boolean flag with an enum field.
The existing behavior will return a list of values if childMatch is true.
In the new behavior childMatch can return either a list or a map.
Providing this feature speeds up my job by x5 for its particular
use-case of childMatch by allowing for the deletion of a specific
key without iterating through all the elements.

### DIFF
--- a/hydra-task/src/main/java/com/addthis/hydra/task/output/tree/PathQuery.java
+++ b/hydra-task/src/main/java/com/addthis/hydra/task/output/tree/PathQuery.java
@@ -185,7 +185,7 @@ public final class PathQuery extends PathOp {
                     updated = valueList.updateBundle(state.getBundle());
                     break;
                 default:
-                    throw new IllegalStateException("Known childMatch value " + childMatch);
+                    throw new IllegalStateException("Unknown childMatch value " + childMatch);
             }
             if (updated) {
                 if (debug > 0) {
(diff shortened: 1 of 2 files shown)


tea-dragon:
this looks like a breaking change with respect to existing configs that might try to use a boolean value. Not necessarily a deal breaker, but worth mentioning.

In any event, I think in most cases "2 enum values or null" can be better replaced by "3 enum values". You can even give the enums the method that does this matching logic so that you can just do `childMatch.updateBundle` rather than manual switch logic.

hydra-task/src/main/java/com/addthis/hydra/task/output/tree/PathQuery.java
```diff
@@ -167,10 +174,18 @@ private boolean evaluateNode(TreeMapState state, String[] pathValues, DataTreeNo
...
+                    updated = valueList.updateBundleWithListAppend(state.getBundle());
+                    break;
+                case AS_MAP:
+                    updated = valueList.updateBundleWithMapAppend(state.getBundle());
+                    break;
+                case NONE:
+                    updated = valueList.updateBundle(state.getBundle());
+                    break;
+                default:
+                    throw new IllegalStateException("Known childMatch value " + childMatch);
```
tea-dragon: probably supposed to be "Unknown" rather than "Known"?

PR #142, opened 2015-01-23
Complete filter initialization is now explicit. Ref T51853.

mspiegel:
Remaining instances where new filter initialization needs to be invoked. Replaced BundleField with AutoFields in data attachments. In cases where the field name is only serialized in the configuration object, we can replace the String type with the AutoField type. In cases where we are serializing the field name in the data attachment itself (the binary representation) we probably shouldn't be doing this but it is out of scope to fix that behavior so we preserve the type as a String and use AutoField.createNewField().


mspiegel:
This PR passes the nightly test suite.

PR #143, opened 2015-01-25
Implement 'maxNodes' to PathValue and derivatives.

mspiegel:
Optionally specify the maximum number of children nodes that can be created in a subtree. Several one line changes to remove redundant code.

This is a fairly small change in terms of LoC but it affects one of the core components of Hydra jobs so it is more appropriate as a pull request.

### DIFF
--- a/hydra-task/src/main/java/com/addthis/hydra/task/output/tree/PathValue.java
+++ b/hydra-task/src/main/java/com/addthis/hydra/task/output/tree/PathValue.java
@@ -90,6 +90,15 @@ public class PathValue extends PathElement {
     @FieldConfig(codable = true)
     protected PathElement each;
 
+    /**
+     * If positive then limit the number of nodes that can be created.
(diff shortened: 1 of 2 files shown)


hydra-task/src/main/java/com/addthis/hydra/task/output/tree/PathValue.java
```diff
@@ -150,19 +159,30 @@ public final TreeNodeList getNextNodeList(final TreeMapState state) {
...
             return TreeMapState.empty();
         }
-        TreeNodeList list = null;
+        TreeNodeList list;
         if (sync) {
             synchronized (this) {
                 list = processNodeUpdates(state, value);
             }
-        } else if (value != null) {
+        } else {
```
mspiegel: ValueUtil.isEmpty(value) on line 147 tests for null.

tea-dragon:
non-racey version would be cool, but lgtm I suppose.

PR #144, opened 2015-01-27
Preempt prune/deletion operations

mspiegel:
Added 'preempt' boolean parameter to PathPrune. Allows preemption to be terminated when process is shutting down. Moved background node deletion in the ConcurrentTree from currently at the end of a job as a process that is not preempted to the new behavior of the beginning of a job as a process that is preempted when the JVM exits. 

Changes included:
    
- Moved the invocation of process() out of the TreeMapState constructor. This is also not necessary for the new feature it has just bugged me a lot.
    
- Added atomic boolean field 'closing' to TreeMapper class. Can be tested by other classes to determine if JVM is shutting down.
    
- Added boolean field 'preempt' to PathPrune with default value of false. Use 'preempt' in combination with TreeMapper#closing to determine if preemption is activated.

### DIFF
--- a/hydra-data/src/main/java/com/addthis/hydra/data/tree/DataTree.java
+++ b/hydra-data/src/main/java/com/addthis/hydra/data/tree/DataTree.java
@@ -15,6 +15,8 @@ package com.addthis.hydra.data.tree;
 
 import java.io.IOException;
 
+import java.util.function.BooleanSupplier;
+
 import com.addthis.hydra.store.db.CloseOperation;
 
 public interface DataTree extends DataTreeNode {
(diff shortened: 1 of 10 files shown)


tea-dragon:
- don't think the changes to StreamMapper open/close are needed
- not a huge fan of having both `close` and `sendComplete`. In this case, `close` doesn't even actually release any resources, even though there are resources that need closing. It also looks like it is not expected for the owning object to invoke it.

mspiegel:
Yeah I convinced myself that the StreamMapper open/close was just thread-safety theater it wasn't actually doing anything. It has been removed. I'll take a look at the close and sendComplete in the morning.

mspiegel:
This pull request is ready for feedback and pending any feedback it can be merged into master.

PR #145, opened 2015-02-01
remove [BV]Filter.open and make into interfaces

tea-dragon:
also removes VFilter.acceptNull because it seems way preferable
to perform a few extra `if (value == null) { return null; }`
checks during a VFChain than to make people handle the otherwise
terribly confusing special casing involved.


mspiegel:
LGTM

PR #146, opened 2015-02-07
long node ids

mspiegel:
64-bit node identifiers have been implemented. Keys in the database are serialized using a variable length encoding of the node identifiers. Identifiers less than or equal to Integer.MAX_INT are encoded in 4 bytes. Identifiers greater than Integer.MAX_INT are encoded in 8 bytes and the most significant bit of the first byte is set as a flag. Keys in a page use delta-encoding from the first key of the page to reduce disk space. Existing hydra jobs can be converted to the new representation. Serialization is managed by the enumerated type PageEncodeType. The interface KeyCoder is the best place to look at the new API. The class has been documented extensively.

This pull request requires https://github.com/addthis/codec/pull/11.


mspiegel:
The latest version of this pull request 07ff901f23a18dd2f4b78fba97438f2d1f31a568 has passed the tests in the test cluster.

PR #150, opened 2015-02-20
Implement maximum over-estimation for top-K.

mspiegel:
This is an extension of https://github.com/addthis/hydra/pull/148. Added maximum over-estimation for top-K. Feature is optional and disabled by default. If enabled then each top-K item tracks the maximum over-estimation of its associated count. Query results will return this error as an entry in a map data attachment. Implemented option to query only gauranteed top-K items which is convenient. Documentation for top-K data attachment has been updated.

### DIFF
--- a/hydra-data/src/main/java/com/addthis/hydra/data/tree/concurrent/ConcurrentTreeNode.java
+++ b/hydra-data/src/main/java/com/addthis/hydra/data/tree/concurrent/ConcurrentTreeNode.java
@@ -375,7 +375,7 @@ public class ConcurrentTreeNode extends AbstractTreeNode {
         return true;
     }
 
-    HashMap<String, TreeNodeData> createMap() {
+    public HashMap<String, TreeNodeData> createMap() {
         if (data == null) {
             data = new HashMap<>();
         }
--- a/hydra-data/src/main/java/com/addthis/hydra/data/util/KeyTopper.java
+++ b/hydra-data/src/main/java/com/addthis/hydra/data/util/KeyTopper.java
@@ -63,6 +63,9 @@ public final class KeyTopper implements Codable, BytesCodable {
     @FieldConfig(codable = true)
     private boolean lossy;
 
+    @FieldConfig(codable = false)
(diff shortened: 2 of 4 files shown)


hydra-data/src/main/java/com/addthis/hydra/data/tree/concurrent/ConcurrentTreeNode.java
```diff
@@ -375,7 +375,7 @@ public boolean aliasTo(DataTreeNode node) {
         return true;
     }
 
-    HashMap<String, TreeNodeData> createMap() {
+    public HashMap<String, TreeNodeData> createMap() {
```
tea-dragon: let's push this visibility change down to virtual tree node. data has a complicated enough life cycle as it is, and it already causes contention where it shouldn't be required so it is on my short list for fixes if we start to have perf problems related to that.

don't want to risk it becoming even messier if we don't have to.

hydra-data/src/main/java/com/addthis/hydra/data/util/KeyTopper.java
```diff
@@ -63,6 +63,15 @@ public KeyTopper() {
     @FieldConfig(codable = true)
     private boolean lossy;
 
+    /**
+     * Error estimates are only supported in the BytesCodable
+     * serialization format. They are not supported
+     * in the older serialization format to preserve
+     * serialization compatibility.
+     */
+    @FieldConfig(codable = false)
```
tea-dragon: I believe it would suffice to omit the annotation from a programmatic perspective. I am inclined to think the javadoc by itself would also be easier to understand since it does not require a mental diff on the annotation lines.

tea-dragon:
not exactly thrilled with all the required nullity checks and serialization tricks, but it takes up no extra disk space and barely any more memory (when it isn't turned on) so I can't really complain. Not much that can be done for most of that without eg. moving off bytes codable anyway. At least modern codec supports manual ordering of fields so that adding a field with a name that isn't alphabetically last doesn't break everything.

PR #151, opened 2015-02-20
Improve error handling of WriteableDiskCheck.

mspiegel:
Addresses the issues that were brought up in https://github.com/addthis/hydra/pull/131


tea-dragon:
In the past, I have found extending thread to be a fantastically bad idea unless absolutely necessary. I think this one can be easily replace with a Task. That said, I'm not convinced we need to pull out concurrency anyway. `touch` shouldn't be throwing out enough data to risk filling a pipe buffer, so we can probably just read it afterwards. I think I solved this same problem previously in `EmailUtil.java` for the `mailx` shell-out.. maybe take a look there?

mspiegel:
I took a look at EmailUtil. As you said it doesn't consume from standard output and it only consumes from standard error if an error occurred. I am not a fan of expecting the output buffers to be small where the failure mode is that the process doesn't complete. I guess it is mitigated by the timeout in the process waitFor() which is good. On the other hand this error handling does cause the subprocess to not complete. Which for EmailUtil means that an email is not delivered and the only notification we have is to happen to search through the log messages and notice "Oops this important email was never delivered."

I did some refactoring to my first implementation. The StreamGobbler has been replaced by a StandardOutputError thread that consumes both standard output and standard error. There's maybe a solution for this problem that uses NIO and avoids the need for the consuming thread. Have at it I didn't know how to make that.

tea-dragon:
The implementation in EmailUtil was an incremental improvement on what was there before, so some things like triggering more severe warnings or handling hanging sub processes were a little out of scope. The key feature it provided was being an obvious improvement over completely dropping error information or hanging forever waiting for the sub process to complete (which also makes it hard to report errors).

In any event, how the error is reported is less important than ensuring it is detected to begin with. For writable disk check, it has meters and its return value.

I am still not comfortable with creating one or more threads on every invocation, and still don't think it is necessary to be pro-active about ensuring the buffer doesn't fill up. If we time out on our waitFor, we can simply close stderr to unblock the process, (wait), sigterm it, (wait), sigkill it. It is unlikely to end up with so much stderr output being produced for this to matter, but we can definitely be quite certain that even if it is, we don't care about it.

Additionally, I am not sure that this implementation that attempts to use one thread to read from both streams will always work as expected. If reads block when the stream is empty, and a process writes to only stderr, it will still end up being blocked. If reads don't block, then it's going to immediately exit.

mspiegel:
Fair points on the implementation. I may teach myself some NIO to drain the
buffers without a background thread.

mspiegel:
Actually I think NIO is overkill. I can probably put together a simpler
solution.

tea-dragon:
I don't think those NIO calls are actually non-blocking.

tea-dragon:
- I think there might have been a good reason I checked both the exit value and stderr contents. There could have been a failure type that returned 0 but printed some stderr message.

- it looks like we will no longer get any warning or log message if the send email shell-out does not complete. The writable disk check will fail by returning false, but it won't say why -- that could not happen before.

nice to haves, but not necessarily required to be incrementally better:
- some form of clean up after time outs
- clearer api "hasCompleted" kind of makes it seem like it might change if you wait a while
- probably just throw exceptions when a bad thing happens. email util can catch them if it really wants

hydra-data/src/main/java/com/addthis/hydra/util/EmailUtil.java
```diff
@@ -13,11 +13,7 @@
  */
 package com.addthis.hydra.util;
 
-import java.io.OutputStreamWriter;
-
-import java.util.concurrent.TimeUnit;
-
-import com.google.common.io.ByteStreams;
+import com.addthis.basis.util.Strings;
```
tea-dragon: need to update to master for new basis namespaces

PR #152, opened 2015-02-27
Implement pipeline job type

mspiegel:
A pipeline job will run the first job phase to completion, then the second phase, then the third phase, etc. When a pipeline job stops and is restarted it begins processing from the first phase. Added error checking to verify that each job phase does not write to an output directory of another job phase. The support for conflicting output directories can be used to backport this feature to ordinary hydra jobs with conflicting mark dirs or conflicting outputs. Ref T53265.

### DIFF
--- a/hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeeder.java
+++ b/hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeeder.java
@@ -103,20 +103,23 @@ public final class MapFeeder implements Runnable {
         }
 
         try {
+            boolean interrupt = false;
             if (source.isEnabled()) {
                 while (fillBuffer()) {
                     if (Thread.interrupted()) {
+                        interrupt = true;
                         closeSourceIfNeeded();
                     }
                 }
             }
+            final boolean interrupted = interrupt;
             closeSourceIfNeeded();
             joinProcessors();
             log.info("all ({}) task threads exited; sending taskComplete", feeders);
 
             // run in different threads to isolate them from interrupts. ie. "taskCompleteUninterruptibly"
             // join awaits completion, is uninterruptible, and will propagate any exception
-            CompletableFuture.runAsync(task::taskComplete).join();
+            CompletableFuture.runAsync(() -> task.taskComplete(interrupted)).join();
             // critical to get any file meta data written before process exits
             CompletableFuture.runAsync(MuxFileDirectoryCache::waitForWriteClosure).join();
         } catch (Throwable t) {
--- a/hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeeder.java
+++ b/hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeeder.java
@@ -115,7 +115,7 @@ public final class MapFeeder implements Runnable {
             log.info("all ({}) task threads exited; sending taskComplete", feeders);
 
             // run in different threads to isolate them from interrupts. ie. "taskCompleteUninterruptibly"
-            // join awaits completion, is uninterruptible, and will propogate any exception
+            // join awaits completion, is uninterruptible, and will propagate any exception
             CompletableFuture.runAsync(task::taskComplete).join();
             // critical to get any file meta data written before process exits
             CompletableFuture.runAsync(MuxFileDirectoryCache::waitForWriteClosure).join();
--- a/hydra-task/src/main/java/com/addthis/hydra/task/map/StreamMapper.java
+++ b/hydra-task/src/main/java/com/addthis/hydra/task/map/StreamMapper.java
@@ -124,8 +124,8 @@ public class StreamMapper implements StreamEmitter, TaskRunnable {
     private final LongAdder outputTime = new LongAdder();
 
     // metrics
-    private final Meter inputMeter = Metrics.newMeter(getClass(), "input", "input", TimeUnit.SECONDS);
-    private final Meter outputMeter = Metrics.newMeter(getClass(), "output", "output", TimeUnit.SECONDS);
+    private static final Meter inputMeter = Metrics.newMeter(StreamMapper.class, "input", "input", TimeUnit.SECONDS);
+    private static final Meter outputMeter = Metrics.newMeter(StreamMapper.class, "output", "output", TimeUnit.SECONDS);
 
     @GuardedBy("metricGate") private long lastTick;
     @GuardedBy("metricGate") private long lastOutputTime = 0;
--- a/hydra-task/src/main/java/com/addthis/hydra/task/map/StreamMapper.java
+++ b/hydra-task/src/main/java/com/addthis/hydra/task/map/StreamMapper.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 
 import java.net.ServerSocket;
 
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.LongAdder;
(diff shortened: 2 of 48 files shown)


mspiegel:
Confirmed that task metrics reported to Spawn are calculated as the aggregate numbers across all phases of a pipeline job. This behavior was working due to a bug in the StreamMapper metric fields. This bug has been fixed with correct behavior maintained in 11a2057af4e88101820f717bc3bada39afbf497f.

hydra-task/src/main/java/com/addthis/hydra/task/pipeline/PipelineTask.java
```diff
@@ -0,0 +1,170 @@
...
+     */
+    private volatile boolean phaseIsRunning = false;
+
+    @JsonCreator
+    public PipelineTask(@JsonProperty("phases") @Nonnull StreamMapper[] phases,
+                        @JsonProperty("validateDirs") boolean validateDirs) {
+        this.phases = phases;
+        this.validateDirs = validateDirs;
+        this.manager = new Thread(new PipelineTaskManager(), "PipelineTask");
+        String message = validateOutputDirectories();
```
tea-dragon: seems like this method could just be void and throw the exception itself

hydra-task/src/main/java/com/addthis/hydra/task/pipeline/PipelineTask.java
```diff
@@ -0,0 +1,170 @@
...
+    @Override public void start() {
+        manager.start();
+    }
+
+    @Override public void close() throws Exception {
+        manager.interrupt();
+        manager.join();
+    }
+
+    @Override public void emit(Bundle bundle) {
```
tea-dragon: who is calling this method?

tea-dragon:
Error handling for the "manager" thread seems less than robust. It looks like it might end up with a lot of uncaught exceptions and I doubt that will do good things.

tea-dragon:
Directory validation seems heavy weight for what it provides -- especially without any default implementation. Even then it would be much more appealing if it either seemed easy to extend to solve similar problems or involved modifying less interfaces (eg. had its own interface).

mspiegel:
Test and verified that exceptions thrown in either the init() or close() methods will cause the job to error. If an exception occurs during a phase in the pipeline then no downstream phases are executed and the job fails.

hydra-task/src/main/java/com/addthis/hydra/task/map/StreamMapper.java
```diff
@@ -366,4 +375,27 @@ private void maybeCloseJmx() {
             }
         }
     }
+
+    public ImmutableList<String> outputRootDirs() {
+        return ImmutableList.<String>builder()
+                .addAll(source.outputRootDirs())
+                .addAll(output.outputRootDirs())
+                .build();
+    }
```
tea-dragon: this can be an implementation of your interface, right?

hydra-task/src/main/java/com/addthis/hydra/task/output/OutputRootDirectories.java
```diff
@@ -0,0 +1,33 @@
...
+public interface OutputRootDirectories {
+
+    /**
+     * List of root directories that are created. Some data sources
+     * write persistent state to a directory in addition to
+     * reading from other sources. Each implementing
+     * class is responsible for specifying an URL scheme when
+     * not referring to a path on the filesystem (ie. hdfs:// for example).
+     */
+    public default @Nonnull ImmutableList<String> outputRootDirs() {
```
tea-dragon: these can probably be `Path`s then instead of strings

hydra-task/src/main/java/com/addthis/hydra/task/pipeline/PipelineTask.java
```diff
@@ -0,0 +1,154 @@
...
+    private final boolean validateDirs;
+
+    private final CompletableFuture<Void>[] phaseManagement;
+
+    private volatile StreamMapper currentPhase = null;
+
+    @JsonCreator
+    public PipelineTask(@JsonProperty("phases") @Nonnull StreamMapper[] phases,
+                        @JsonProperty("validateDirs") boolean validateDirs) {
+        validateOutputDirectories();
```
tea-dragon: you can't call this before `this.validateDirs = validateDirs;`. It will always appear to be false. I am a little surprised that even compiles.

hydra-task/src/main/java/com/addthis/hydra/task/map/StreamMapper.java
```diff
@@ -302,7 +306,7 @@ public void close() throws InterruptedException {
     }
 
     /** called on process exit */
-    public void taskComplete() {
+    public void taskComplete(boolean interrupted) {
```
tea-dragon: the map feeder is only ever interrupted on the stream mapper's say-so. There is no need to pass a flag from stream mapper to the map feeder via interrupting the thread and then pass that flag back to stream mapper from map feeder by adding it this signature.

hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeeder.java
```diff
@@ -103,20 +103,23 @@ public MapFeeder(StreamMapper task, TaskDataSource source, int feeders) {
...
+            boolean interrupt = false;
             if (source.isEnabled()) {
                 while (fillBuffer()) {
                     if (Thread.interrupted()) {
+                        interrupt = true;
                         closeSourceIfNeeded();
                     }
                 }
             }
+            final boolean interrupted = interrupt;
```
tea-dragon: it looks like there could be a race condition where the map feeder gets interrupted during many of the later steps after it runs out of bundles in the normal manner. eg. source is exhausted, loop exits, map feeder starts to close sources etc, (this takes a little while), interrupt is sent, map feeder finishes exiting without seeing interrupt.

hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeeder.java
```diff
@@ -103,20 +103,23 @@ public MapFeeder(StreamMapper task, TaskDataSource source, int feeders) {
...
             joinProcessors();
             log.info("all ({}) task threads exited; sending taskComplete", feeders);
 
             // run in different threads to isolate them from interrupts. ie. "taskCompleteUninterruptibly"
-            // join awaits completion, is uninterruptible, and will propogate any exception
-            CompletableFuture.runAsync(task::taskComplete).join();
+            // join awaits completion, is uninterruptible, and will propagate any exception
+            CompletableFuture.runAsync(() -> task.taskComplete(interrupted)).join();
             // critical to get any file meta data written before process exits
             CompletableFuture.runAsync(MuxFileDirectoryCache::waitForWriteClosure).join();
```
tea-dragon: this looks like this can get called after another task in the pipeline has started, but this method acts on a static instance that may be being used by the new task. Aside from angry log messages about being unable to close things, I think it will eventually try to force close files / directories and that will probably cause errors for the next task.

tea-dragon:
if it's all the same to you, I think I'd like to see the new task type and the output dir cross-field validation logic as two different PRs. Each one is large enough on its own

hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeeder.java
mspiegel: That's fine I would consider that a benign race condition. There is no harm done if the onTaskComplete fires. It just means we could open a source then is immediately closed.

tea-dragon: I'm not sure I follow. It looks like being interrupted implies someone called close, and it looks like close is only ever called on one sub-task. How is the new source getting immediately closed?

hydra-task/src/main/java/com/addthis/hydra/task/map/StreamMapper.java
mspiegel: I'm doing this wrong. I'll fix it.

mspiegel:
Closing in favor of https://github.com/addthis/hydra/pull/156. Don't delete this branch yet. I have to split it into the 2nd pull request.

PR #153, opened 2015-03-03
Added empty output aliased to FilteredDataOutput. Ref T53231

yuesong:
Instead of adding a new class, I opted for an alias to FilterDataOutput with empty chain filter and an empty tree output.

I wanted to use unary (true) in extra-filters, but that will require jobs to include extra-filters. An empty chain was the best thing I could come up with.

### DIFF
--- a/hydra-task/src/main/resources/reference.conf
+++ b/hydra-task/src/main/resources/reference.conf
@@ -86,6 +86,10 @@ plugins {
     ganglia: GangliaOutput
     http: DataOutputHttp
     tree: tree.TreeMapper
+    empty: { _class: filtered
+      filter: []
+      output: {tree {root:[{const:"EMPTY"}]}}
+    }
   }
 
   output-factory {


tedpearson:
Any reason not to consider moving unary into reference.conf to support this feature?

tea-dragon:
What is the purpose of the filtered data output wrapper?

I think I would also prefer if such an output did not entail writing files or constructing a tree mapper (for testing purposes).

yuesong:
Uh.. you are right. I was looking for examples and saw FilteredDataOutput. That somehow stuck with me. Could've been an alias to an empty tree mapper. You seem to not like that.  For testing, it's just as good, no? Or did I miss something.

tea-dragon:
It is very nearly as good, but tree mapper has tons of settings and log output. It would also be nice if it could be excluded as a potential problem while debugging an error.

yuesong:
Fair enough. I'll close this one and submit another PR.

PR #154, opened 2015-03-04
Added EmptyDataOutput. Ref T53231

yuesong:


tea-dragon:
in an instance of complete inconsistency and unfairness, while sources don't need to be thread safe, I believe outputs actually do. Not super important for this use case, but might want to document that bundle count accordingly.

tea-dragon:
lgtm, think we might want to eventually give it another name though. empty makes more sense for a source than an output. then again it doesn't make a ton of sense for our input given that it emits rows at all...

PR #155, opened 2015-03-04
Added DataSourceConstant. Ref T53234

yuesong:


hydra-task/src/main/java/com/addthis/hydra/task/source/DataSourceConstant.java
```diff
@@ -0,0 +1,90 @@
...
+            val = Bundles.deepCopyBundle(val);
+        }
+        return val;
+    }
+
+    private Bundle currentBundle() {
+        return (cursor >= bundles.length || cursor < 0)? null : bundles[cursor];
+    }
+
+    private void advanaceCursor() {
```
tea-dragon: typo here. advanAce should be advance

tea-dragon:
looks reasonable, although you might get some mileage out of some methods from `Iterators`:

`Iterators.limit`, `Iterators.concat`, `Iterators.cycle`, `Iterators.peekingIterator`

yuesong:
Thanks for the suggestion. peek and next changed to throwing NoSuchElementException instead of returning null

PR #156, opened 2015-03-04
Implementation of pipeline job type. Ref T53265.

mspiegel:


hydra-task/src/main/java/com/addthis/hydra/task/map/StreamMapper.java
```diff
@@ -366,4 +370,9 @@ private void maybeCloseJmx() {
             }
         }
     }
+
+    public CompletableFuture<Void> getOnTaskComplete() {
```
tea-dragon: Maybe we should call it completionFuture or something. "onTaskComplete" more describes a specific use case (adding hooks) than what it is. It might be kinda awkward looking to do `task.getOnTaskComplete().join()` for instance.

hydra-task/src/main/java/com/addthis/hydra/task/pipeline/PipelineTask.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override public void close() throws Exception {
+        log.info("Pipeline task is starting shutdown");
+        int size = phaseComplete.size();
+        for (int i = (size - 1); i >= 0; i--) {
+            phaseComplete.get(i).cancel(false);
+        }
+        for (int i = (size - 1); i >= 0; i--) {
+            try {
+                if (!phaseComplete.get(i).isCancelled()) {
```
tea-dragon: can we use the result from the cancel attempt earlier?

hydra-task/src/main/java/com/addthis/hydra/task/pipeline/PipelineTask.java
```diff
@@ -0,0 +1,137 @@
...
+                    phaseNext.get(i).join();
+                }
+            } catch (CompletionException ex) {
+                log.error("Phase {} phaseNext future encountered an exception while starting phase {}: ",
```
tea-dragon: logging and then rethrowing errors is usually not a great sign, but the indexing information does seem helpful. Maybe this is a use case for wrapping exceptions -- eg. `new SomeExceptionType("Phase .... while starting phase...").

PR #157, opened 2015-03-05
Query input source for reading local trees. Ref T53270.

mspiegel:
This implementation can be used to use a query as an input source when the tree is within the same job. It is intended for pipeline jobs.


hydra-task/src/main/java/com/addthis/hydra/task/source/DataSourceQuery.java
```diff
@@ -0,0 +1,279 @@
...
+        this.ops = ops;
+        this.fields = fields;
+        this.buffer = buffer;
+        this.pollInterval = pollInterval;
+        this.pollCountdown = pollCountdown;
+        this.queue = new LinkedBlockingQueue<>(this.buffer);
+    }
+
+    @Override public void init() {
+        localInputSource = Objects.equals(System.getenv("HYDRA_JOBID"), jobId);
```
tea-dragon: alternatives:
- have job id be optional and have null imply local source (can also rename it to "remoteJobId")
- check that the data directory is a descendant of the working directory
- make another field called "localJobId" with the appropriate default and check against that

hydra-task/src/main/java/com/addthis/hydra/task/source/DataSourceQuery.java
```diff
@@ -0,0 +1,279 @@
...
+     */
+    @Nullable private final String[] fields;
+
+    /**
+     * Maximum size of the queue that stores bundles prior to their processing.
+     * Default is 128.
+     */
+    private final int buffer;
+
+    @Time(TimeUnit.MILLISECONDS)
```
tea-dragon: has to go with the constructor parameter

hydra-task/src/main/java/com/addthis/hydra/task/source/DataSourceQuery.java
```diff
@@ -0,0 +1,279 @@
...
+     * Operations for the query. This field is optional.
+     */
+    @Nullable private final String ops;
+
+    /**
+     * Optionally specify the field names for the columns
+     * produced by the query. Otherwise the fields will
+     * be named "0", "1", "2", etc. Default is null.
+     */
+    @Nullable private final String[] fields;
```
tea-dragon: should just be `AutoField[]`

hydra-task/src/main/java/com/addthis/hydra/task/source/DataSourceQuery.java
```diff
@@ -0,0 +1,279 @@
...
+            closeFuture.join();
+        }
+    }
+
+    @Override public void send(Bundle input) {
+        try {
+            Bundle output = input;
+            if (fields != null) {
+                output = output.createBundle();
+                for (int i = 0; i < fields.length; i++) {
```
tea-dragon: I think iterating over sequential index fields is the same as doing `for (BundleField field : bundle.getFormat())`

hydra-task/src/main/java/com/addthis/hydra/task/source/DataSourceQuery.java
```diff
@@ -0,0 +1,279 @@
...
+            }
+        } finally {
+            if (interrupted) {
+                Thread.currentThread().interrupt();
+            }
+        }
+    }
+
+    @Override public Bundle peek() throws DataChannelError {
+        return queue.peek();
```
tea-dragon: not allowed to return null from peek before any bundles are produced or you know they never will be

hydra-task/src/main/java/com/addthis/hydra/task/source/DataSourceQuery.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    @Override public void close() {
+        if (shuttingDown.compareAndSet(false, true)) {
+            try {
+                if (queryHandle != null && queryHandle.isAlive()) {
```
tea-dragon: calling Thread.isAlive() seems like a bad sign. I don't think fixing "query engine source single" or equivalent should be considered out of scope here

mspiegel:
I think I've incorporated all the feedback from the first round but I haven't yet addressed the query engine single source. I created the abstract class AbstractStreamDataSource that is now shared by AbstractStreamFileDataSource and DataSourceQuery.  Make comments on the new version and I'll include those changes in the next commit and a fix for query engine single source.

hydra-task/src/main/java/com/addthis/hydra/task/source/DataSourceQuery.java
```diff
@@ -0,0 +1,192 @@
...
+        super.init();
+        localInputSource = (jobId == null);
+        if (!localInputSource) {
+            throw new UnsupportedOperationException("At this time only local queries are supported");
+        }
+        try {
+            File dir = Paths.get(treeDir).toFile();
+            tempDir = Files.createTempDir();
+            queryEngine = new QueryEngineSourceSingle(new ReadTree(dir));
+            Query query = new Query(System.getenv("HYDRA_JOBID"), new String[] {path}, null);
```
tea-dragon: jobId field over `System.getenv("HYDRA_JOBID")`?

mspiegel: I think jobId is null for local queries which is the only thing implemented
at the moment.

tea-dragon: then you can have another field or a TaskRunConfig object. Directly accessing environment variables something we would like to avoid. It is the equivalent of a static field

mspiegel:
Closing in favor of more general query input source pull request.

PR #160, opened 2015-03-11
Added support for querying one or more specified tasks. Ref T53414

yuesong:


hydra-main/src/main/java/com/addthis/hydra/query/web/HttpQueryCallHandler.java
```diff
@@ -80,6 +80,7 @@ public static ChannelFuture handleQuery(ChannelHandler queryToQueryResultsEncode
         }
         query.setParameterIfNotYetSet("allocator", kv.getValue("allocator"));
         query.setParameterIfNotYetSet("allowPartial", kv.getValue("allowPartial"));
+        query.setParameterIfNotYetSet("tasks", kv.getValue("tasks"));
```
tea-dragon: If it's going to get first class treatment in the UI, maybe we should move this statement up with the other A listers. And/Or use a real field rather than a dynamic parameter.

tea-dragon:
Consistency between UI fields and full Query field support (logging, ui display, etc) would be nice, but this looks good to me.

tea-dragon:
Ah, shoot. This doesn't handle task count validation. Oh well, wasn't critical for an initial version.

PR #161, opened 2015-03-12
LeasedTreeNodeList abstraction

mspiegel:
Add LeasedTreeNodeList abstraction.
    
The purpose is to simplify the node leasing contract by moving all the logic into a central location. The collection will ensure that either its contents are read exactly once or it is released exactly once. The guarantees are weak in that if the library developer forgets to invoke release we don't do anything other then issue some logging messages and fail on an assertion and these actions happen in a finalizer (bleh).

Added node.release() to the end of TreeMapState#pop() and removed return value from TreeMapState#pop(). We always release the node that is popped and then drop the reference to the node so the API should reflect this pattern.


hydra-task/src/main/java/com/addthis/hydra/task/output/tree/LeasedTreeNodeList.java
```diff
@@ -0,0 +1,61 @@
...
+    protected LeasedTreeNodeList(List<DataTreeNode> nodes) {
+        super(nodes);
+    }
+
+    @Override protected void releaseItem(DataTreeNode data) {
+        data.release();
+    }
+
+    /**
+     * Should we delete this method is it worthless?
```
tea-dragon: Yes, we should delete it. It isn't worthless, but its cost far exceeds our its value. Creating possibly hundreds of thousands of finalizers a second is really, really not where we want to be.

hydra-task/src/main/java/com/addthis/hydra/task/output/tree/ReadOnceList.java
```diff
@@ -0,0 +1,78 @@
...
+
+import javax.annotation.concurrent.NotThreadSafe;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Spliterator;
+import java.util.function.Consumer;
+
+/**
+ * Collection who contents can either be read exactly once or
```
tea-dragon: "Collection" might be misleading here since it does not extend the actual Collection class, and it is a relatively important/ common class, and it has many of the same methods.

hydra-task/src/main/java/com/addthis/hydra/task/output/tree/ReadOnceListLeakDetection.java
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Collection who contents can either be read exactly once or
+ * whose contents can be released
+ * @param <T>
+ */
+@NotThreadSafe
+public class ReadOnceListLeakDetection<T> implements ReadOnceList<T> {
+
+    @Nonnull
+    protected ResourceLeak leak;
```
tea-dragon: can these be final? I can see they are protected, so maybe you have a use case in mind where they change, but having trouble finding it

tea-dragon:
I have to say I am finding the semantics of the read-once list to kinda confusing. There also seem to be plenty of cases where it won't actually detect leaks, and while I understand that those cases mostly mirror what existed before, I think an argument could be made that they were more obviously possible flaws.

Now a user might, fairly reasonably, attempt to rely on the API methods to validate that things are handled correctly.

I have a suggestion, which is to do leak detection on tree nodes directly. Hiding some of the boilerplate for handling leases / null / etc is nice, but maybe we should just move those to library methods. The reasoning is that users already still have to worry about tree node leases, so we might as well let them use a list interface they are probably familiar with (ie. List).

mspiegel:
I talked to Ian. We can close this PR in favor of a simpler approach.

PR #162, opened 2015-03-16
Hydra job query input source.

mspiegel:
Input source will either use the mqworkers to run a query
or run the query directly if no job identifier is provided.
Spawn or mqmaster are not used when running queries only the
mqworker network.

Refactors the AbstractStreamFileDataSource to extract the
functionality for streaming data sources into AbstractStreamDataSource.

At present no type of mark directory is implemented for the
query input source until we determine what is appropriate
for this input source.

Ref T53270. Ref T53475.


mspiegel:
Don't delete branch yet. I will cannibalize it.

PR #163, opened 2015-03-17
Kafka8

stevedonnelly:
Implements v1 of Kafka 8 hydra source.  Also includes a variety of wrappers around kafka consumer and producer api calls.


hydra-kafka/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+      <groupId>com.addthis.hydra</groupId>
+      <artifactId>hydra-task</artifactId>
+    </dependency>
+
+
+      <!-- 3rd party -->
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-clients</artifactId>
+      <version>0.8.2.0</version>
```
tea-dragon: should probably put this version in dependency management or at least a property (since it is repeated and apparently likely to change a few times)

stevedonnelly: Agreed, and done.

mspiegel:
Will be moved to https://github.com/addthis/hydra-contrib

PR #165, opened 2015-03-19
POC for better work stealing in MapFeeder. Ref T53505

yuesong:
This is a proof of concept of using interrupt to notify main map feeder thread of hungry worker threads. When a worker thread drained its queue, it interrupts the main thread, which then batch fill all the drained queues before switching back to the normal round-robin-like dispatching.

It appears to work well in the case of one worker thread progressing much slower than the other, which is the main use case for this optimization. In the general case where filtering and output are slower than source, the code should have very low to no overhead. However, in the case of source being the bottleneck, the worker threads will interrupt the map feeder thread a lot for no benefit. Suggestions on how to improve that is welcome. There are other lesser optimizations to do, e.g. preventing too many interrupted exceptions etc. but I want to get some feedback on the general approach before fine tuning.

Also I repurposed existing "task.worksteal" parameter for this, assuming the current work stealing implementation isn't being used. If that's not the case, it can be easily changed to a new parameter.

### DIFF
--- a/hydra-main-api/src/main/java/com/addthis/hydra/task/source/TaskDataSource.java
+++ b/hydra-main-api/src/main/java/com/addthis/hydra/task/source/TaskDataSource.java
@@ -14,7 +14,6 @@
 package com.addthis.hydra.task.source;
 
 import com.addthis.bundle.channel.DataChannelSource;
-import com.addthis.bundle.core.BundleField;
 import com.addthis.codec.annotations.FieldConfig;
 import com.addthis.codec.annotations.Pluggable;
 import com.addthis.codec.codables.Codable;
@@ -38,7 +37,7 @@ public abstract class TaskDataSource implements Codable, DataChannelSource, Clon
      * hash function to determine which input processing thread to use. Default is null.
      */
     @FieldConfig(codable = true)
-    private BundleField shardField;
+    private String shardField;
 
     /** If false then disable this data source. Default is true. */
     @FieldConfig(codable = true)
@@ -46,7 +45,7 @@ public abstract class TaskDataSource implements Codable, DataChannelSource, Clon
 
     public abstract void init();
 
-    public final BundleField getShardField() {
+    public final String getShardField() {
         return shardField;
     }
 
--- a/hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeeder.java
+++ b/hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeeder.java
@@ -13,6 +13,7 @@
  */
 package com.addthis.hydra.task.map;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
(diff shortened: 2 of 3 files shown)


hydra-main-api/src/main/java/com/addthis/hydra/task/source/TaskDataSource.java
```diff
@@ -38,7 +37,7 @@
      * hash function to determine which input processing thread to use. Default is null.
      */
     @FieldConfig(codable = true)
-    private BundleField shardField;
+    private String shardField;
```
mspiegel: This should be an AutoField instead of a String.

hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeeder.java
```diff
@@ -134,27 +141,75 @@ private void handleUncaughtThrowable(Throwable t) {
...
             }
-            int mod = Math.abs(hash % queues.length);
-            pushQueue(mod, p);
-            return true;
-        } catch (NoSuchElementException ignored) {
-            log.info("exiting on premature stream termination");
         }
-        return false;
+        if (!interruptedBundlePutBack) {
+            log.error("Batch fill is called when no queue is drained. This is a bug!");
```
mspiegel: If this error will lead to data being processed incorrectly then we should throw an IllegalStateException and have the job error.

mspiegel:
I think the following convention predated the pull request but it would be nice to cleanup if we are making changes to the class. There is a field "private final int feeders" but there is only one feeder thread. It looks like this variable is storing the number of mapper tasks. It should be renamed.

mspiegel:
MapFeeder#steal() is no longer used. It should be deleted.

mspiegel:
I think we should remove the shardField entirely. I don't see its purpose. There is no local state stored in each MapperTask (unless there is some dark corner of hydra using ThreadLocal variables that I don't know about) which implies that the selection of which MapperTask should not have an impact on the processing of bundles.

mspiegel:
I'm not a fan of using Thread interruption as a signalling mechanism for coordination across threads. We should be able to accomplish the same task using some of the data structures in the java.util.concurrent package. Also this pull request isn't implementing work stealing. When a mapper task runs out of work it signals the producer thread and waits for more work to become available. In a work stealing implementation it would take elements from the other queues.

yuesong:
Thanks for the feedback. What I really want to focus on is the best coordination mechanism. I think using interruption minimizes locks (on data structures). What do you think may be a better alternative?

yuesong:
Turns out interrupting the main thread will cause some sources to close prematurely. Michael's implementation using ForkJoinPool works pretty well in my test, so I'm closing this pull request and will open another one.

tea-dragon:
+1. I don't think interruption is anybody's favorite these days

PR #166, opened 2015-03-20
Query notation for specifying a default value.

mspiegel:
There was a feature request to implement regular expression type
matching in the query path syntax for repeated expressions of
type X{n}, X{n,}, X{n,m}. This is a simpler form of that feature.
It implements a query path notation of the form "/+[foo]" or
"/+[foo]a,b,c". The string enclosed in the square brackets is a
default value that is returned when no matches are found. The
same notation works for the data attachment queries "/+%[foo]name=parameters"

There are several benefits to this approach. It is simpler
than the original request and was implemented quickly. It has
a more straightforward semantics in the sense that it ensures that
a column is always generated for the "+" request. It was unclear
whether there would be rows with variable-length columns in the
original X{n,m} request. It also meets the use case of the original
feature request.

Ref T52644.

### DIFF
--- a/hydra-data/src/main/java/com/addthis/hydra/data/query/QueryElementNode.java
+++ b/hydra-data/src/main/java/com/addthis/hydra/data/query/QueryElementNode.java
@@ -38,6 +38,7 @@ import com.addthis.hydra.data.tree.ReadTreeNode;
 import com.addthis.hydra.data.tree.TreeNodeData;
 import com.addthis.hydra.data.tree.prop.VirtualTreeNode;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.AbstractIterator;
 import com.google.common.collect.Iterators;
 
@@ -205,7 +206,8 @@ public class QueryElementNode implements Codable {
      * the input token. Optionally "+[text|hits]" can be used to set
      * the number of hits on the default node.
      */
-    private String extractDefaultValue(String tok) {
+    @VisibleForTesting
+    String extractDefaultValue(String tok) {
         int startDefault = -1;
         if (tok.startsWith(DEFAULT_NODE)) {
             startDefault = DEFAULT_NODE.length();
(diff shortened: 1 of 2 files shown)


hydra-data/src/main/java/com/addthis/hydra/data/query/QueryElementNode.java
```diff
@@ -188,6 +199,39 @@ public QueryElementNode parse(String tok, MutableInt nextColumn) {
...
+                defaultValue = tok.substring(startDefault, hitsLocation);
+            } else {
+                defaultValue = tok.substring(startDefault, endDefault);
+            }
+            if (tok.startsWith(DEFAULT_NODE)) {
+                tok = "+" + tok.substring(endDefault + 1);
+            } else if (tok.startsWith(DEFAULT_ATTACHMENT)) {
+                tok = "+%" + tok.substring(endDefault + 1);
+            } else {
+                throw new IllegalStateException("Unexpected state in default value extraction");
```
tea-dragon: doesn't look like this can be reached as a result of the internal state this method accesses (defaultHits, defaultValue). Therefore, should probably be an illegal argument exception?

hydra-data/src/main/java/com/addthis/hydra/data/query/QueryElementNode.java
```diff
@@ -188,6 +199,39 @@ public QueryElementNode parse(String tok, MutableInt nextColumn) {
         return this;
     }
 
+    /**
+     * The prefix "+[text]" or "+%[text] is interpreted as a default
+     * value. Extract the default value and remove it from
+     * the input token. Optionally "+[text|hits]" can be used to set
+     * the number of hits on the default node.
+     */
+    private String extractDefaultValue(String tok) {
```
tea-dragon: method is.. pretty intense. Given that it is called only once per query (well, per worker maybe?), I think you should feel free to trade performance for clarity if possible. The docs are pretty great though, so I would also be happy with just a unit test.

tea-dragon:
Things get a little hairier later on, and it is true that we already have quite a few op-based methods of setting default values for things. It might be simpler to just fix the weird "allow empty columns" type deal, but hey if it works, then I can live with it under the assumption that some day the whole class was going to burn in fire anyway.

lgtm

PR #167, opened 2015-03-24
Added MapFeederForkJoin and a StreamMapper flag to use it. Ref T53505

yuesong:
In my local synthetic tests that simulates a few corner cases, the new MapFeederForkJoin achieved better bundle processing throughput than the current impl. In a test where I copied real source data and job config, it did the same.

### DIFF
--- a/hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeederForkJoin.java
+++ b/hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeederForkJoin.java
@@ -1,0 +1,160 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.addthis.hydra.task.map;
+
+
+import java.util.NoSuchElementException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.RecursiveAction;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import com.addthis.basis.util.Parameter;
+
+import com.addthis.bundle.core.Bundle;
+import com.addthis.hydra.task.source.TaskDataSource;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class MapFeederForkJoin implements Runnable {
+    private static final Logger log = LoggerFactory.getLogger(MapFeederForkJoin.class);
+
+    private static final int QUEUE_DEPTH = Parameter.intValue("task.queue.depth", 100);
+
+    // state control
+    private final AtomicBoolean errored = new AtomicBoolean(false);
+    private boolean hasClosedStreams = false; // not shared with MapperTasks
+
+    // enclosing task
+    private final StreamMapper task;
+    private final TaskDataSource source;
+
+    // mapper task controls
+    private final ForkJoinPool mapperPool;
+    private final int parallelism;
+    private final Semaphore enqueuePermits;
+
+    private long start = System.currentTimeMillis();
+    private long totalBundles = 0;
+
+    public MapFeederForkJoin(StreamMapper task, TaskDataSource source, int parallelism) {
+        this.source = source;
+        this.task = task;
+        this.parallelism = parallelism;
+        this.mapperPool = new ForkJoinPool(parallelism);
+        this.enqueuePermits = new Semaphore(QUEUE_DEPTH);
+    }
+
+    @Override public void run() {
+        log.info("starting {} thread(s) for src={}", parallelism, source);
+        try {
+            if (source.isEnabled()) {
+                while (fillBuffer()) {
+                    if (Thread.interrupted()) {
+                        closeSourceIfNeeded();
+                    }
+                }
+            }
+            closeSourceIfNeeded();
+            joinProcessors();
+            log.info("all ({}) task threads exited; sending taskComplete", parallelism);
+            logBundleThroughput();
+            // run in different threads to isolate them from interrupts. ie. "taskCompleteUninterruptibly"
+            // join awaits completion, is uninterruptible, and will propagate any exception
+            CompletableFuture.runAsync(task::taskComplete).join();
+        } catch (Throwable t) {
+            logBundleThroughput();
+            handleUncaughtThrowable(t);
+        }
+        log.debug("task feeder exited");
+    }
+
+    private void logBundleThroughput() {
+        long elapse = (System.currentTimeMillis() - start) / 1000;
+        log.info("{} bundles processed in {} seconds (avg rate={}/s)", totalBundles, elapse, totalBundles / elapse);
+    }
+
+    /**
+     * Immediately halting is the best way we have of ensuring errors are reported without a significantly
+     * more involved system.
+     */
+    private void handleUncaughtThrowable(Throwable t) {
+        if (errored.compareAndSet(false, true)) {
+            log.error("unrecoverable error in task feeder or one of its mapper threads. immediately halting jvm", t);
+            Runtime.getRuntime().halt(1);
+        }
+    }
+
+    /**
+     * Push the next bundle onto the work queue. Returns true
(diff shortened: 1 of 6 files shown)


cburroughs:
:heart: forkjoin

tea-dragon:
gotta give you points for craftyness, and forkjoin is one of those weird places where it is implemented so much more efficiently than a lot of its friends that even if you are neither forking nor joining a single time, sometimes (it at least feels) more efficient.

without comment on overall design or efficiency, I have a couple low hanging fruit guys though. i'll just inline them

hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeederForkJoin.java
```diff
@@ -0,0 +1,160 @@
...
+    private final Semaphore enqueuePermits;
+
+    private long start = System.currentTimeMillis();
+    private long totalBundles = 0;
+
+    public MapFeederForkJoin(StreamMapper task, TaskDataSource source, int parallelism) {
+        this.source = source;
+        this.task = task;
+        this.parallelism = parallelism;
+        this.mapperPool = new ForkJoinPool(parallelism);
```
tea-dragon: you can hand the constructor an uncaught exception handler to save on a lot of try/finally

hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeederForkJoin.java
```diff
@@ -0,0 +1,160 @@
...
+    private boolean fillBuffer() {
+        boolean status = false;
+        try {
+            enqueuePermits.acquire();
+            Bundle p = source.next();
+            if (p == null) {
+                log.info("exiting on null bundle from {}", source);
+            } else {
+                totalBundles++;
+                mapperPool.submit(new MapperTask(p));
```
tea-dragon: do we need to both create a task and then have the task do its own post-complete accounting? with the uncaught exception handler and a listener on the future, it can just be like a method reference

hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeederForkJoin.java
```diff
@@ -0,0 +1,160 @@
...
+            log.info("exiting on premature stream termination");
+        } catch (InterruptedException ignored) {
+            Thread.currentThread().interrupt();
+            status = true;
+        }
+        return status;
+    }
+
+    private void joinProcessors() {
+        mapperPool.shutdown();
```
tea-dragon: have to actually then wait for termination

hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeederForkJoin.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Immediately halting is the best way we have of ensuring errors are reported without a significantly
+     * more involved system.
+     */
+    private void handleUncaughtThrowable(Throwable t) {
+        if (errored.compareAndSet(false, true)) {
+            log.error("unrecoverable error in task feeder or one of its mapper threads. immediately halting jvm", t);
+            Runtime.getRuntime().halt(1);
```
tea-dragon: new method in basis will replace this

hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeederForkJoin.java
mspiegel: Whoops. And we probably want to wait for a finite amount of time and error the job if the wait does not complete successfully.

tea-dragon:
even if the tests are admittedly synthetic, it would be cool to share them

hydra-task/src/main/java/com/addthis/hydra/task/map/MapFeederForkJoin.java
yuesong: what method?

tea-dragon: it's in the latest snapshot, and I mean specifically the logic contained in the handler (the Runtime.getRuntime().halt(1) call). The current namespacing is `Shutdown.exit(int)`

tea-dragon: (I originally used halt there because Runtime.exit(int) is so unreliable and crazy)

yuesong:
As for the synthetic tests, they are simple jobs using constant source (which introduces imbalanced work sharding). I created three test cases, for each of which I run three jobs with identical config except that: one uses the original MapFeeder, one uses the forkjoin MapFeeder (`useForkJoinMapFeeder=true`), and one uses the original MapFeeder with work stealing (`// -Dtask.worksteal=true`), then I just compare the final throughputs. The test configs (using the original MapFeeder) are

- full speed (no filtering, empty output)
```
// -Dtask.worksteal=false

include "extra-filters"

source.const{
    bundles: [
        // thread 0
        {SHARD:"foo", A:a}
        // thread 1
        {SHARD:"bar", B:b}
        {SHARD:"bar", B:b}
        {SHARD:"bar", B:b}
        {SHARD:"bar", B:b}
    ]
    shardField:"SHARD"
    repeat:10000000
}

map.fields:[SHARD, A, B]
map.filterOut:[]

output.empty {}
```
- tree output (no filtering, simple tree output)
```
// -Dtask.worksteal=false

include "extra-filters"

source.const{
    bundles: [
        // thread 0
        {SHARD:"foo", A:a}
        // thread 1
        {SHARD:"bar", B:b}
        {SHARD:"bar", B:b}
        {SHARD:"bar", B:b}
        {SHARD:"bar", B:b}
    ]
    shardField:"SHARD"
    repeat:5000000
}

map.fields:[SHARD, A, B]
map.filterOut:[]

output.tree.root=[
    {const:root}
    {field:SHARD}
]
```
- slow filter (some bundles take longer than others) - this is the main use case we want to optimize
```
// -Dtask.worksteal=false

include "extra-filters"

source.const{
    bundles: [
        // thread 0
        {SHARD:"foo", A:a}
        // thread 1
        {SHARD:"bar", B:b}
        {SHARD:"bar", B:b}
        {SHARD:"bar", B:b}
        {SHARD:"bar", B:b}
    ]
    shardField:"SHARD"
    repeat:10000
}

map.fields:[SHARD, A, B]
map.filterOut:[
    {from:SHARD, to:HASH, hash {}}
    {log:sharding, maxBundles:10}
    {safely:[
        {has:A}
        {sleep.duration:2}
    ]}
    {safely:[
        {has:B}
        {sleep.duration:1}
    ]}
]

output.empty {}
```

mspiegel:
It may be out of scope for this pull request but we can remove the shardField from the TaskDataSource. Unless there is some secret ThreadLocal variables there is no purpose for having a worker thread process a specific partition of the data.

tea-dragon:
while I am by no means in favor of the shard field, there are regrettably a non-trivial number of possible uses. I haven't seen anyone try these or anything but:

- manipulation of the queues as seen in the tests (to somehow influence priority or downstream behavior)
- some filters have transient, thread-safe logic that can be keyed on bundles. For instance, all AutoField objects would prefer to have bundles of the same format from the same thread -- they don't use actual thread locals, but they use non-volatile fields that inherently benefit from cpu caching
- other filters have transient or even non-transient (builders et al) logic that is sometimes keyed on certain bundle fields

the most obvious one though is outputs. They always have to handle concurrent threads, and usually have side effects that care about more than their current bundle (and almost always including the bundles processed by other threads). You can imagine how sharding on a certain field that partitions a tree fairly well might improve throughput.

tea-dragon:
As for the test cases, it would be relatively easy to make them into test classes right? Even just as a gist or something. These cases are nice, but they do seem a little biased to wards cases that are better for this implementation, and the sharding adds quite a few unknowns of its own.

What if the source is being the bottleneck? What is the change in CPU/memory overhead? For instance, one possibly worse case I can imagine is heavy cpu contention on the system and especially in the source (expensive decompression/deser perhaps). 

My gut says those costs are much less than eg. getting stuck even if we were to run into that case much more frequently (thus why I made the ticket), but performance can be a very unintuitive thing sometimes and usually worth considering all the angles.

yuesong:
I'm not familiar with how sharding bundles to different worker threads can affect tree output efficiency? Regardless of shardFiled, as far as the output is concerned, the bundles are being added in somewhat random order with respect to the sharding, right? For implementation like the forkjoin one, it isn't applicable. I guess I just don't see people understanding/using the shardField, so...

The reason for the lack of a test case where source is the bottleneck is really that I can't easily introduce slowness to a source. That reminds me of my plan of using a real source for this...

yuesong:
I just did a quick test of using a mesh2 source with empty map and output (so 97% of time is spent in source). Original MapFeeder had the lowest throughput, while MapFeederForkJoin and the original with work stealing performed the same.

tea-dragon:
To be clear, I was specifically worried about a case where system CPU was constrained. Unless you had a million threads and were monitoring the overall system load, I don't see how this way measured.

PR #168, opened 2015-04-02
Stop rabbitmq queues from overflowing

mspiegel:
This pull request has a minion deregister itself from the 'ALL' routing key when it shuts down. The existing implementation has a minion register itself to 'ALL' each time it starts up. This pull request does not delete the rabbitmq queues. After some discussion it was decided not to implement deletion of the queues. If spawn was to delete the queue then the minion would immediately shutdown when the queues were deleted. We could change the behavior of the minion but then we run into the possibility of a minion running disconnected to a queue. If the minion is to delete the queue when it shuts down this circumvents the durable property of rabbit queues and we are not certain if something is relying on the durability.

Two downsides of this approach: (1) the queues for a dead host no longer receive messages but they will slowly build up over time. In practice this happens very slowly. (2) If a minion receives a SIGKILL it will not deregister itself and the queues will continue to build up. Proper monitoring of the rabbitmq will be adequate to manually delete the queues in this condition.

### DIFF
--- a/hydra-mq/src/main/java/com/addthis/hydra/mq/RabbitMessageConsumer.java
+++ b/hydra-mq/src/main/java/com/addthis/hydra/mq/RabbitMessageConsumer.java
@@ -13,6 +13,8 @@
  */
 package com.addthis.hydra.mq;
 
+import javax.annotation.Nonnull;
+
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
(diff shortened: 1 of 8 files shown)


hydra-mq/src/main/java/com/addthis/hydra/mq/RabbitMessageConsumer.java
```diff
@@ -86,9 +100,24 @@ public void handleDelivery(String consumerTag, Envelope envelope, AMQP.BasicProp
     }
 
     @Override public void close() throws IOException {
+        IOException firstError = null;
```
tea-dragon: see: suppressed exceptions

PR #169, opened 2015-04-03
Spawn quiesces if rabbitmq cannot emit messages.

mspiegel:
The rabbitmq network enters an unhappy state if messages are being produced and not consumed. But the cluster is in a very bad state when messages can no longer be produced. When messages are dropped on the floor the cluster can enter an inconsistent state. This automatically quiesces the cluster to minimize the damage. Ref T54274.
    
This pull request branches from https://github.com/addthis/hydra/pull/168

### DIFF
--- a/hydra-mq/src/main/java/com/addthis/hydra/mq/MeshMessageProducer.java
+++ b/hydra-mq/src/main/java/com/addthis/hydra/mq/MeshMessageProducer.java
@@ -50,7 +50,7 @@ public class MeshMessageProducer implements MessageProducer {
 
     private MessageFileProvider provider;
 
-    public MeshMessageProducer(final MeshyClient mesh, final String topic) {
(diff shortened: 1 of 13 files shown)


hydra-mq/src/main/java/com/addthis/hydra/mq/MeshMessageProducer.java
```diff
@@ -60,8 +60,13 @@ public MeshMessageProducer(final MeshyClient mesh, final String topic) {
         }
     }
 
-    @Override
-    public void open() throws IOException {
+    public static MeshMessageProducer constructAndOpen(MeshyClient client, String topic) throws IOException {
```
tea-dragon: needs javadoc saying the returned thing must be closed

tea-dragon:
some of the comments on the other one apply here as well. they look at least more reasonable than before so fine with merging spose

PR #170, opened 2015-04-16
codec hocon docs Ref T52379

tedpearson:
Documentation on how to simplify jobs from the old JSON format to the new HOCON format. @tea-dragon please review and correct

### DIFF
--- a/docs/source/tutorial/codec.rst
+++ b/docs/source/tutorial/codec.rst
@@ -45,8 +45,7 @@ identifiers and javascript-style comments, both illegal in JSON.
 
 JSON is actually a subset of HOCON, which means that any JSON hydra job is actually also a HOCON job. HOCON also
 supports the non-quoted identifiers and comments that our "JSON" job format supported. So if you were
-to ask "How do I convert a JSON job to HOCON" the answer is actually "Your job is already HOCON"! However, Hydra
-only enables the HOCON parser if a job does not have curly braces at the start and end.
+to ask "How do I convert a JSON job to HOCON" the answer is actually "Your job is already HOCON"!
 
 Some key differences:
 ---------------------
@@ -54,6 +53,7 @@ Some key differences:
 * Single quoted strings are not supported in HOCON, while they were in the JSON format. In HOCON, a single quoted
   string will result in the string including the single quote characters.
 * Comments in HOCON can be made using either "//" or "#". Our JSON format only allowed "//".
+* The JSON format allowed multi-line comments using "/* comment */" syntax. HOCON does not have multi-line comments.
 * There is only one additional reserved keyword in HOCON: "include". You must not use "include" anywhere in a job config
   except in quoted strings, unless using it for its file include function.
 * You must escape strings as in java. In the JSON format, you could do something like "\\" or '"', which would now need
--- a/docs/source/tutorial/codec.rst
+++ b/docs/source/tutorial/codec.rst
@@ -54,10 +54,13 @@ Some key differences:
   string will result in the string including the single quote characters.
 * Comments in HOCON can be made using either "//" or "#". Our JSON format only allowed "//".
 * The JSON format allowed multi-line comments using "/* comment */" syntax. HOCON does not have multi-line comments.
-* There is only one additional reserved keyword in HOCON: "include". You must not use "include" anywhere in a job config
-  except in quoted strings, unless using it for its file include function.
+* There is only one additional reserved keyword in HOCON: "include". You should not use "include" unquoted
+  in a job config, unless using it for its file include function. However it will work in a quoted string.
+  If you happen to have a parameter called "include" you can avoid errors by using quotes:
+  `{my-plugin."include":"some_value"}`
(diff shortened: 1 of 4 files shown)


docs/source/tutorial/codec.rst
```diff
@@ -0,0 +1,365 @@
...
+the only job format supported was the JSON format.
+
+JSON format jobs are actually somewhat of a misnomer. Most JSON jobs that we wrote internally at AddThis were not
+actually "JSON" in the strict sense. They were more of a javascript format, as Codec 2 supported non-quoted
+identifiers and javascript-style comments, both illegal in JSON.
+
+JSON is actually a subset of HOCON, which means that any JSON hydra job is actually also a HOCON job. HOCON also
+supports the non-quoted identifiers and comments that our "JSON" job format supported. So if you were
+to ask "How do I convert a JSON job to HOCON" the answer is actually "Your job is already HOCON"! However, Hydra
+only enables the HOCON parser if a job does not have curly braces at the start and end.
```
tea-dragon: hasn't been true for a while. we always use the hocon parser now

docs/source/tutorial/codec.rst
```diff
@@ -0,0 +1,365 @@
...
+
+Some key differences:
+---------------------
+* Values in HOCON don't have to be quoted. For most production hydra jobs, our convention is to use quotes anyways.
+* Single quoted strings are not supported in HOCON, while they were in the JSON format. In HOCON, a single quoted
+  string will result in the string including the single quote characters.
+* Comments in HOCON can be made using either "//" or "#". Our JSON format only allowed "//".
```
tea-dragon: we allowed multi-line comments before as well (those aren't allowed anymore)

docs/source/tutorial/codec.rst
```diff
@@ -0,0 +1,365 @@
...
+* There is only one additional reserved keyword in HOCON: "include". You must not use "include" anywhere in a job config
```
tea-dragon: I think it is fine as long as it isn't a key or isn't unquoted

docs/source/tutorial/codec.rst
```diff
@@ -0,0 +1,365 @@
...
+  except in quoted strings, unless using it for its file include function.
+* You must escape strings as in java. In the JSON format, you could do something like "\\" or '"', which would now need
```
tea-dragon: the "\\" example would parse/ validate, but it was not actually working correctly in almost every case

docs/source/tutorial/codec.rst
```diff
@@ -0,0 +1,365 @@
...
+   _uidcount {
+       uid.count:{key:"UID", max:50000000}
+       uidage.count:{key:"UID_GOOD", max:50000000}
+       uidage48.count:{key:"UID_SUPER_GOOD", max:50000000}
+   }
+
+   output.tree.root: [
+      {const:"ROOT"}
+      {branch:[
+         [
```
tea-dragon: these nested arrays are pointless. might as well remove them to provide a better example

docs/source/tutorial/codec.rst
tedpearson: That's what the following line says. Is that good enough?

docs/source/tutorial/codec.rst
tedpearson: I'm not sure how to put that in "parsable by a job writer" syntax, can I just leave it out?

tea-dragon: maybe? I would mention that character escaping might a source of job validation errors if converting some very old job config, but that in that case, it is most likely that the old config is doing something unexpected / wrong.

docs/source/tutorial/codec.rst
tedpearson: I don't understand how the arrays are pointless. An explanation would be helpful.

docs/source/tutorial/codec.rst
tea-dragon: "must not use "include" anywhere in a job config" sounded a little overly restrictive as an introduction to the keyword. im sure it is fine though.

docs/source/tutorial/codec.rst
tedpearson: Added

docs/source/tutorial/codec.rst
tea-dragon: as in, removing the nested arrays would result in the same job config. The default mode of branch is an array of sibling nodes. Putting each sibling node inside an array triggers the default "array" logic of path elements... that is, a parent->child relationship. Since those arrays are size one, they accomplish nothing.

tea-dragon: eg.

`output.tree.root {const: ROOT}` would work, but
`output.tree.root: [ ... ]` uses the default path element as an array handling to create a list of parent->child layers.

That same logic is being used for your nested arrays, and they are pointless for the same reason `output.tree.root: [ {const: ROOT} ]` would be

tedpearson: In other words, The only reason you need nested arrays is if you want each sibling to be more than one node deep. Got it.

PR #171, opened 2015-04-28
fix broken link to ValueFilter list page

ranadessr:


tedpearson:
You missed a "hydra" after "addthis" in the fix.

PR #173, opened 2015-05-06
Add Spawn#unsafeGetJob() endpoint.

mspiegel:
For use cases when reading a stale copy of a job is OK. With 7b2a9119ad1b5a70f1c960e770d1e33112588775 the known deadlock in Spawn should be fixed. We still have the potential for deadlock in future code as the Spawn#getJob() method acquires the job lock. Which is probably a a good thing except in cases when it's ok to read a stale copy of the job. Such as alerts. The downside of this pull request is that it introduces extra complexity and allows people to bypass the job lock. The job lock may be removed at some point but for now we should respect it (except in those cases we don't need to?)


mspiegel:
Can reopen later if needed.

PR #176, opened 2015-06-23
Fixed range docstring to correspond to code

mythguided:
Docstring says it does from a to b, but code pops last, then first, then string. Fixed docstring to correspond to code.


tea-dragon:
lgtm

PR #177, opened 2015-07-10
changing spawn & minion command messages to use (Codec) json

stevedonnelly:
Original purpose of these changes was to add some addition fields (owner and group) to kick commands for resource usage tracking.  Ended up making a variety of changes/clean-up so this sort of thing would be easier in the future.

Note, since the serialization format for command messages is changing, this would cause any persisted messages in the old format to be be dropped.  This should not be a problem for us since the standard release process clears all non-repeating messages from the queue.

Note, I haven't finished testing yet, but I wanted to start the review as soon as possible so this can make it into Monday's release.

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/job/mq/CommandTaskNew.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/mq/CommandTaskNew.java
@@ -13,17 +13,18 @@
  */
 package com.addthis.hydra.job.mq;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, defaultImpl = CommandTaskNew.class)
 public class CommandTaskNew extends AbstractJobMessage {
 
-    private static final long serialVersionUID = 1117109955580588580L;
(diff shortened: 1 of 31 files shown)


hydra-main/src/main/java/com/addthis/hydra/job/mq/CommandTaskNew.java
```diff
@@ -16,14 +16,12 @@
 
 public class CommandTaskNew extends AbstractJobMessage {
 
-    private static final long serialVersionUID = 1117109955580588580L;
+    @SuppressWarnings("unused")
+    public CommandTaskNew() {
```
tea-dragon: constructors that exist only for codec / jackson should be either protected or private. That'll probably also handle the warning.

tea-dragon:
looks pretty okay overall. Still a few warts, bu a pretty clear improvement at least.

abramsm:
I appreciate the detailed commit message, very helpful.

mspiegel:
I think I will submit a patch in the near future that will replace all the if/else if/else if/ that use instanceof with dynamic dispatch, since that's the entire point of dynamic dispatch. But this pull request looks good enough for now.

stevedonnelly:
Testing completed.  Had to default type info annotations to handle several cases where messages (specifically HostState and CommandTaskKick) were being serialized to disk/zookeeper as part of Spawn/Minion state.  Rebased on master afterwards.

PR #178, opened 2015-07-23
Optionally disable the node cache in ConcurrentTree

abramsm:
Performance testing has indicated that the node cache (MediatedEvictionConcurrentHashMap) may actually hinder performance.  This PR creates a new parameter which allows a user to optionally disable this cache.  Further testing will decide if we should make this the default case or not


tea-dragon:
I am definitely interested in having this as an option and in further testing, but the implementation adds some complexity in undesirable places. For instance, I would very much like to kill the current magic system properties that control tree settings (and cannot be done per tree) rather than add to them.

At a glance, I'd suggest the cleanest way to add this is as an alternate tree implementation, rather than a boolean switch to this existing one.

abramsm:
closing this in favor of a new PR that totally removes the cache

PR #179, opened 2015-08-02
Remove node cache from DataTree to improve performance

abramsm:
testing indicates that map jobs run faster without the cache.  This version of code passes all unit tests but has not been through a complete hercules build.  Hoping to merge this into master and then get the automated analysis from that.


abramsm:
closing this for now.  @mspiegel pointed out a potential concurrency issue.  all tests are passing but we believe an edge case failure is possible

PR #184, opened 2015-08-20
Provide a unsafe tree implementation (non-concurrent) for improved single thread performance

abramsm:
Hydra jobs that use a single thread still pay a high price for concurrency in Hydra.  This PR provides an optional path that uses unsafe data structures which should allow for improved performance in single thread operation. 

The default behavior of jobs with this PR will be exactly the same as they are today.  If the job configuration sets threads=1 and concurrentTree=false then the non-concurrent codebase will be used.  If this proves successful we may consider making non concurrent the default behavior and allow job developers to move to a concurrent environment on an as needed basis.

My testing in hercules has shown that most CPU bound 2 threaded jobs are out performed by single threaded non concurrent configurations.  Jobs with a heavy amount of IO tend to perform better with a multi threaded configuration.

Comments welcome.


tea-dragon:
The changes to abstract tree node do not work well with ReadTreeNode. The other pre-existing subclass.

abramsm:
I meant to go back and look at  ReadTree but forgot.  not play well as in functional issue or implementation?

tea-dragon:
The added fields throw off cache sizing, and neither they nor the added methods have any valid function for it.

mspiegel:
Oops yes we need to be preserve correctness of the ReadTree or else the query system memory estimation will be incorrect and the query system will OOM.

abramsm:
K, good catch.  I'll take a look and see if I can correct unless some beats
me to it.

mspiegel:
I did some minor cleanup with respect to logging and node deletion in both the concurrent tree and non-concurrent tree. While I was doing that I noticed that the non-concurrent tree performs only foreground deletion of nodes. This makes some sense as the tree doesn't allow for concurrent deletion. The disadvantage of the nonconcurrent tree implementation is that deleting a subtree does not stop when the tree is closed, which is undesirable if the task runtime has been exceeded or we are quiescing the cluster for maintenance.

I think the correct solution is to have the noncurrent tree move nodes into the trash node and then have the #foregroundNodeDeletion() method responsible for cleaning up the trash.

abramsm:
@mspiegel yeah, thats a good point on shutdown concerns.  I was hoping to avoid the overhead of the trash node.  i'll take a look but i wonder if there is a solution to solve for the blocking shutdown without requiring a trash root.

abramsm:
@tea-dragon d97da05e256976071a2cf8adac61c83de68ab240 should address the ReadTreeNode issue.  I just pushed those methods down.

mspiegel:
Can you update your IDE settings to prevent "import *'s". I've committed the code style changes to the branch.

abramsm:
Yeah, i was actually int he process of doing that and lost my train of thought.  will do prior to next commit

mspiegel:
This pull request looks ready to merge IMO

tea-dragon:
Now seems as good as time as any, although I admit I would be more hesitant if any implied future work wasn't automatically considered worthwhile.

PR #185, opened 2015-08-20
Add backoff param to JSONFetcher and related ValueFilters

alstaplesmoore:


mspiegel:
I did the wrong thing earlier. For this example you want the callable to go ahead and throw the IOException and not wrap it in an UncheckedIOException.

alstaplesmoore:
I retry on UncheckedIOException. My adhoc tests show retries working correctly on an invalid URL.

mspiegel:
@alstaplesmoore can you take a look at this change: https://github.com/addthis/hydra/commit/862887c17c0ee400c76168cc6ae0c1441bbd9871? My request is that we do not wrap IOExceptions inside of UncheckedIOExceptions when it's not necessary.

alstaplesmoore:
^ Looks fine

PR #186, opened 2015-08-20
added int-to-enum conversion methods

gnurik:
These are useful for http client that looks at an int value and tries to convert to the enum


mspiegel:
I think the method you want is #makeState(int value) and it is already implemented.

gnurik:
ah yes, don't know how I missed that, thx

PR #187, opened 2015-09-01
Remove MediatedEvictionConcurrentHashMap.

mspiegel:
Use a regular ConcurrentHashMap. release() operations that return the final lease are responsible for eviction to disk. The node remains in the cache with a lease count of -1 this is used as a tombstone
in the cache while the eviction to disk is taking place.

Thanks to an earlier change where nodes must be leased prior to deletion, I don't think there is the possibility of a race condition with node deletion and node eviction. Also note that node state "-3" has been removed in this branch. It is no longer needed.

In the performance tests this branch performance no worse and sometimes slightly better than the MediatedEvictionConcurrentHashMap. Results are https://goo.gl/yk2ztF (internal access only).


tea-dragon:
Is the implication of this that the node cache will always miss unless there is enough contention between threads?

mspiegel:
Err. Performance numbers show a 10x reduction in performance for one class of critical jobs.

tea-dragon:
Closing for now, as it does not seem safe to become the default (and only) option at this time. Might make an interesting toggle.

PR #189, opened 2015-09-28
Implement task-based runtime exceeded alert.

mspiegel:
Added start time field to JobTask object. I don't visit Spawn town very often. Did I miss something?

### DIFF
--- a/hydra-api/src/main/java/com/addthis/hydra/job/JobTask.java
+++ b/hydra-api/src/main/java/com/addthis/hydra/job/JobTask.java
@@ -54,6 +54,7 @@ public final class JobTask implements Codable, Cloneable, Comparable<JobTask> {
     @FieldConfig private int state;
     @FieldConfig private int runCount;
     @FieldConfig private int starts;
+    @FieldConfig private long startTime;
     @FieldConfig private int errors;
     @FieldConfig private long fileCount;
     @FieldConfig private long fileBytes;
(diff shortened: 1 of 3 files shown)


tea-dragon:
I see we are piling straw onto the camal. This is a change I've always wanted though so we'll let that aspect slide.

A concern and a question:

- are the changes to the task consistently ordered with respect to the reads in the alert check? And if so, is there a potential false positive causing race condition where (state -> busy), (alert sees task as running but with previous start time), (task start time is updated)?

- is this change intended to exclusively catch cases where "explicit" (not rekick timed) kicks somehow reset a job's start time while running tasks still exist? Since the alert still checks both the job and the tasks, it seems like this will only catch cases where the job claims to have been running for a shorter amount of time than one of its tasks. That strikes me as the real problem -- most of the utility of task based runtime alerts is lost when bounded by job time this way.

mspiegel:
- I can reorder the reads for the task status check to match the job status check.
- I don't follow the second comment. This patch is to catch instances where users decide to use the kick:// functionality to rekick a job. The job start time will be refreshed but a task can remain stuck forever and the job level runtime exceeded alert will not fire.

mspiegel:
Hmm I guess we use the value of null for the Job start time to reset it's value. I can do the same thing for the tasks.

mspiegel:
Well actually that is a little silly. I can just use a value of 0 to reset the start time for tasks

tea-dragon:
>This patch is to catch instances where users decide to use the kick:// functionality to rekick a job.

That probably could have gone in the commit message or PR summary, yeah?

>The job start time will be refreshed but a task can remain stuck forever and the job level runtime exceeded alert will not fire.

My point was that the job start time should not be "refreshed" when it has yet to stop. This seems like a very roundabout way to address that expectation mismatch between kick:// and alert behavior.

PR #190, opened 2015-10-13
query maxSimul support

yuesong:


hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
```diff
@@ -55,54 +59,13 @@ public void run() {
...
-                            // This QuerySource does not have this file anymore. Signal to the caller that a retry may resolve the issue.
-                            sourceAggregator.replaceQuerySource(taskSource, taskSource.getSelectedSource());
-                        } else {
-                            // This query source has started sending lines. Need to fail the query.
-                            throw io;
-                        }
-                    }
-                }
-            }
+            // XXX both provider and readFrame update sourceAggregator.completed
```
mspiegel: What does the "XXX" mean?

tedpearson: That's similar to // NOTE:

tedpearson: (Eclipse convention)

tea-dragon: what else would a line comment be.. is it an attention drawing thing? like `!!!!!!!`?

tedpearson: In eclipse it was a highlight keyword. like FIXME or TODO

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/MeshSourceAggregator.java
```diff
@@ -133,12 +133,36 @@ void stopSources(String message) {
...
 
-    boolean tryActivateSource(QueryTaskSourceOption option) {
-        return option.tryActivate(meshy, queryOptions);
+    /**
+     * Attemps to active one of the options of the given QueryTaskSource.
+     *
+     * @param taskSource the QueryTaskSource to activate
+     * @return {@code true} if one of the options is activated; {@code false} if none of the
+     *         options can be/is activated
+     */
```
tea-dragon: how much of this javadoc can be inferred from the method name, parameters, and return type?

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
```diff
@@ -145,7 +182,85 @@ private static AutoField getSourceField(Query query) {
                 throw new UncheckedIOException(e);
             }
         } else {
-            return new NoopField();
+            return null;
+        }
+    }
+
+    private static class FrameReadResult {
```
tea-dragon: Class doesn't appear to be used

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
```diff
@@ -136,6 +150,29 @@ private void rescheduleSelfWithBackoff() {
...
+        } else {
+            return new MaxSimulFrameReadTaskSourceProvider(sourceAggregator, maxSimul);
+        }
+    }
+
+    private int getMaxSimul(@Nullable String maxSimulQueryParam, int totalTasks) {
+        if (maxSimulQueryParam == null) {
+            return totalTasks;
+        }
+        try {
```
mspiegel: Looks like we silently revert to totalTasks if the used entered an invalid value. Shouldn't we error instead?

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/MeshSourceAggregator.java
yuesong: not all

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
yuesong: I think this can go either way - ignoring invalid maxSimul will not cause any problem, other than the fact that task sources are not activated eagerly. I basically opted for less code.

mspiegel: I disagree. If the user provides an invalid input the query should error. Silently ignoring the maxSimul parameter violates the principle of least surprise. Throw an IllegalArgumentException it is not any more lines of code than this implementation.

abramsm: +1

yuesong: Throwing IAE there will cause 500 response code, which is not any better. The correct solution is to validate the parameter in HttpQueryCallHandler and return 400. There are many unvalidated parameters with the same behavior. Feel free to open a ticket to improve our API overall. I consider that out of scope of this ticket though.

mspiegel: Returning a 500 response code is better than silently creating a monster query to eat up cluster resources. What is an example of another parameter with the same behavior?

yuesong: allocator, allowPartial, timeout

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/MeshSourceAggregator.java
tea-dragon: then we should reduce it to that remainder

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/MeshSourceAggregator.java
```diff
@@ -133,12 +133,36 @@ void stopSources(String message) {
...
+        }
+        return false;
     }
 
-    void replaceQuerySource(QueryTaskSource taskSource, QueryTaskSourceOption option) throws Exception {
+    /**
+     * Resets a QueryTaskSource and attempts to replace the previously selected option with a new one.
+     *
+     * @param taskSource
+     * @throws Exception
```
tea-dragon: blank entries here seem pretty unhelpful

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
tea-dragon: those cases don't seem terribly convincing, but in any event, using exceptions for control flow is usually not great anyway. If the intention here is to explicitly allow for malformed inputs, might as well use `Ints.tryParse`.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
```diff
@@ -145,7 +182,74 @@ private static AutoField getSourceField(Query query) {
                 throw new UncheckedIOException(e);
             }
         } else {
-            return new NoopField();
+            return null;
```
tea-dragon: not sure why this was changed since it seems unrelated, but please try to use `@Nullable` annotations where applicable (or newly applicable).

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
```diff
@@ -145,7 +182,74 @@ private static AutoField getSourceField(Query query) {
...
+        }
+    }
+
+    private interface FrameReadTaskSourceProvider {
```
tea-dragon: this seems like it is just the jdk interface `Supplier`. eg. `Supplier<List<QueryTaskSource>>`. Even the method signatures are the same.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/MeshSourceAggregator.java
```diff
@@ -133,12 +133,36 @@ void stopSources(String message) {
...
+     * @return {@code true} if a new option is activated; {@code false} if selected option is {@code null} or the new
+     *         option could not be activated.
+     */
+    boolean replaceQuerySource(QueryTaskSource taskSource) throws Exception {
+        QueryTaskSourceOption option = taskSource.getSelectedSource();
         taskSource.reset();
+        if (option == null) {
```
tea-dragon: why are we now allowing the option to be null here? That seems like an invalid code path that we would rather have made visible by the subsequent NPE. Now it gives the illusion of being handled correctly with this pretend return value that isn't even checked anywhere.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
```diff
@@ -122,6 +85,57 @@ public void run() {
         }
     }
 
+    private int readFrame(List<QueryTaskSource> taskSources, int maxReads) throws Exception {
```
tea-dragon: "readFrame" is kind of a misleading name here. That might be appropriate for FramedDataChannelReader... but that just has `read` for interface purposes.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
```diff
@@ -55,54 +59,13 @@ public void run() {
...
-                            sourceAggregator.replaceQuerySource(taskSource, taskSource.getSelectedSource());
-                        } else {
-                            // This query source has started sending lines. Need to fail the query.
-                            throw io;
-                        }
-                    }
-                }
-            }
+            // NOTE: both provider and readFrame update sourceAggregator.completed
+            List<QueryTaskSource> taskSources = taskSourceProvider.get();
```
tea-dragon: I'm slightly concerned about the amount of list / iterator garbage that this can churn out.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
yuesong: Old code was creating a new value object even injectSource is not specified. The new code only does that when sourceField is not null.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
yuesong: Me too. Suggestions are welcome.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
yuesong: I assumed it's a "frame" because of AggregateConfig.FRAME_READER_READS. I'm fine with rename it to whatever.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
tea-dragon: trivial wrapper objects (eg. value string) aren't particularly high on my hit list at that scale/ longevity, and there's a strong chance the jvm eliminates it as dead code or suppresses the allocation via escape analysis.

tea-dragon: that said, not objecting to the change, just please use the appropriate annotations.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
tea-dragon: I suggest iterating over the array.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
tea-dragon: "readFrames" (plural) is a little closer, but the important parts that aren't included are the things like lazy activation, selecting other hosts, and injecting source names.

hydra-main/src/main/java/com/addthis/hydra/query/aggregate/QueryTask.java
yuesong: I'm not sure how exactly "iterating over the array" should work? One option is to make `list` and `inactive` field variables. The `get` method will update their elements, instead of creating them every time.

yuesong:
if no more comments, i'll merge.

PR #191, opened 2015-10-21
OpPivot uses input arguments as field indices when looking up field. …

yuesong:
…Ref T59372

OpPivot no longer uses BundleColumnBinder and use rolKeys, colKeys and cellKey as field indices to look up fields. I am not sure what the longer term solution to removing the widely used BundleColumnBinder across the code base is, but hopefully this is a good enough fix for OpPivot while I'm figuring that out.


yuesong:
comments?

PR #193, opened 2015-12-03
Remove unused vendor js files

yuesong:
All js files that are not referenced in index.html or pulled in via require.js are removed. Preliminary test of the UI doesn't find any issue. More comprehensive test would be helpful.


cburroughs:
If it makes your life easier you can remove any of these libraries from NOTICE at the same time.

PR #196, opened 2016-01-26
Fixed distribution when point size < sample size

jhorwit2:
Currently, when the distribution has less data points than the reservoir size the distribution is heavily skewed with 0 values. The fix for this is to pass the snapshot a sublist.

### DIFF
--- a/hydra-data/src/main/java/com/addthis/hydra/data/util/KeyPercentileDistribution.java
+++ b/hydra-data/src/main/java/com/addthis/hydra/data/util/KeyPercentileDistribution.java
@@ -15,7 +15,6 @@ package com.addthis.hydra.data.util;
 
 import com.addthis.codec.annotations.FieldConfig;
 import com.addthis.codec.codables.SuperCodable;
-
 import com.yammer.metrics.stats.CodableUniformSample;
 import com.yammer.metrics.stats.Snapshot;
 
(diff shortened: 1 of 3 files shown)


jhorwit2:
I have one question/concern. Typically when you pick random events from a population for a sample you want to do it with the same chance. Currently, once you reach the reservoir size the probability any new points are added to the sample gets lower and lower since r's upper bound grows and grows. You can see that [here](https://github.com/addthis/hydra/pull/196/files#diff-cd2ff94413c4dfe3d85b0bc6a79222feL66).

tea-dragon:
I believe it is this "lower and lower" property that allows the sample to represent uniform selection over the entire life of the stream. If it was bounded by the reservoir size, that would bias it towards more recent/later data.

PS. Generally, these days unit test methods don't use the "test" prefix in their names since the annotation is considered to handle that designation well enough.

tea-dragon:
(LGTM for merging BTW)

jhorwit2:
After some reading I see what we did is [Vitter's](https://en.wikipedia.org/wiki/Reservoir_sampling) algorithm. The only worrisome thing about this algorithm is that used naively can result in inaccurate results. For example, as the stream grows in size your probability for insertion decreases, which means eventually your list becomes basically static. This can drastically reduce your precision, which means on long running jobs this algorithm is not ideal.

jhorwit2:
@tea-dragon i removed prefix and rebased :+1:

PR #198, opened 2016-02-09
squid:S2095 - Resources should be closed

georgekankava:
squid:S2095 - Resources should be closed.
https://dev.eclipse.org/sonar/coding_rules#q=squid%3AS2095
George Kankava

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/minion/LogUtils.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/minion/LogUtils.java
@@ -38,9 +38,10 @@ public final class LogUtils {
     /** Streams task log files from newest to oldest. The returned Stream should be closed. */
     public static Stream<Path> streamTaskLogsByName(JobTask task) throws IOException {
         Path logDir = task.logDir.toPath();
-        return Files.list(logDir)
-                    .filter(path -> Files.isRegularFile(path, LinkOption.NOFOLLOW_LINKS))
-                    .sorted(Collections.reverseOrder());
+        try(Stream<Path> stream = Files.list(logDir)) {
+            return stream.filter(path -> Files.isRegularFile(path, LinkOption.NOFOLLOW_LINKS))
+                         .sorted(Collections.reverseOrder());
+        }
     }
 
     public static Optional<Path> getNthNewestLog(JobTask task, int runsAgo, String suffix) throws IOException {
--- a/hydra-main/src/main/java/com/addthis/hydra/query/MeshQueryMaster.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/query/MeshQueryMaster.java
@@ -18,9 +18,7 @@ import javax.annotation.Nullable;
 
 import java.io.File;
 import java.io.IOException;
-
 import java.net.InetSocketAddress;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -31,11 +29,11 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.addthis.basis.util.LessStreams;
 import com.addthis.basis.util.LessFiles;
 import com.addthis.basis.util.Parameter;
-
 import com.addthis.hydra.data.query.Query;
 import com.addthis.hydra.data.query.QueryException;
 import com.addthis.hydra.query.aggregate.BalancedAllocator;
@@ -50,7 +48,6 @@ import com.addthis.hydra.query.tracker.QueryTracker;
 import com.addthis.hydra.query.tracker.TrackerHandler;
 import com.addthis.meshy.MeshyServer;
 import com.addthis.meshy.service.file.FileReference;
-
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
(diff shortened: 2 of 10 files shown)


hydra-main/src/main/java/com/addthis/hydra/minion/LogUtils.java
```diff
@@ -38,9 +38,10 @@
     /** Streams task log files from newest to oldest. The returned Stream should be closed. */
     public static Stream<Path> streamTaskLogsByName(JobTask task) throws IOException {
         Path logDir = task.logDir.toPath();
-        return Files.list(logDir)
-                    .filter(path -> Files.isRegularFile(path, LinkOption.NOFOLLOW_LINKS))
-                    .sorted(Collections.reverseOrder());
+        try(Stream<Path> stream = Files.list(logDir)) {
+            return stream.filter(path -> Files.isRegularFile(path, LinkOption.NOFOLLOW_LINKS))
+                         .sorted(Collections.reverseOrder());
+        }
```
tea-dragon: We're returning a Stream here -- it is kind of important that we don't close it prematurely. The javadoc also notes that the caller is responsible for closing it, and most importantly, no (non-fatal) exception can occur that leaves the file stream open. The `IOException` can only occur during the initial `Files.list`, and if it does, then it handles that cleanup internally.

hydra-main/src/main/java/com/addthis/hydra/query/MeshQueryMaster.java
```diff
@@ -246,10 +243,11 @@ protected void writeQuery(ChannelHandlerContext ctx, Query query, ChannelPromise
...
         } else {
-            return LessStreams.stream(TASKS_SPLITTER.split(tasks))
-                              .map(Ints::tryParse)
-                              .filter(i -> i != null)
-                              .collect(Collectors.toSet());
+            try(Stream<String> stream = LessStreams.stream(TASKS_SPLITTER.split(tasks))) {
+                return stream.map(Ints::tryParse)
+                      .filter(i -> i != null)
+                      .collect(Collectors.toSet());
+            }
```
tea-dragon: Unlike `Files.list`, most streams have a no-op close, and the API designers actually did not want obsessive resource protection where it was clearly unneeded. Consider, for example, that `stream.map`, and `.filter` also return `Stream` objects that are resources that could be closed.

Personally, I'm not a fan of the arbitrary "does close matter or not -- read the specifics each time and find out!" kind of API, but unless we want to wrap `map` and `filter` and so on, the line has to be drawn somewhere. A private method with a self-contained stream, backed by a simple `String` is one of the most provably harmless cases there is. I can't see the logic in guarding `.stream` in particular, or any of them for that matter.

hydra-main/src/main/java/com/addthis/hydra/query/web/HttpQueryCallHandler.java
```diff
@@ -65,8 +63,10 @@ public static ChannelFuture handleQuery(ChannelHandler queryToQueryResultsEncode
         if ((dir != null) && !job.endsWith(dir)) {
             String[] jobs = job.split(",");
             String[] dirs = dir.split(",");
-            job = Arrays.stream(jobs).flatMap(subJob ->  Arrays.stream(dirs).map(subDir -> subJob + '/' + subDir))
+            try (Stream<String> stream = Arrays.stream(jobs)) {
+                stream.flatMap(subJob ->  Arrays.stream(dirs).map(subDir -> subJob + '/' + subDir))
                         .collect(Collectors.joining(","));
```
tea-dragon: Same deal as before. Strings to strings to strings. There's no reason to fret over "leaking" the stream any more than there is to to be concerned about `String[] jobs`.

hydra-main/src/main/java/com/addthis/hydra/query/web/HttpStaticFileHandler.java
```diff
@@ -179,35 +179,32 @@ public void channelRead0(
...
             setDateAndCacheHeaders(response, file);
+            if (isKeepAlive(request)) {
+                response.headers().set(CONNECTION, HttpHeaders.Values.KEEP_ALIVE);
+            }
+
+            // Write the initial line and the header.
+            ctx.write(response);
+
+            // Write the content.
+            ctx.write(new DefaultFileRegion(fileChannel, 0, fileLength));
```
tea-dragon: Netty can be kind of a confusing beast, so I'm not surprised your tools got tripped up here. There's a couple issues, some rather significant:

1) the fileChannel should not be closed right after `ctx.write` is called. That write is asynchronous, and it probably won't even try to write at all until that writeAndFlush is called later. In this case, DefaultFileRegion has some error recovery logic where it will construct a new RAF to get a channel from as long as it still has leases left, but that shouldn't be relied on and it breaks the lease paradigm as seen in

2) As soon as we put the fileChannel into that wrapper and give it to `ctx.write`, it is no longer our business to close it. It will be closed in `DefaultFileRegion`'s deallocate method when the latter runs out of leases (pretty much meaning in this case, when it has been fully written to the socket).

3) As soon as `ctx.write(response)` was called, then queuing up the http response header et al, we should strongly prefer to avoid any code paths that could lead to eg. `sendError(ctx, NOT_FOUND)` being called. Most http clients become extremely upset when you send them conflicting http response codes.

4) The separate catch blocks for the two sources of `IOException`s was not my original design (much of this class was appropriated from netty), but I think I would have gone with something not too dissimilar anyway. Of all the exception classes, `IOException` is the most common, and I would even say overused even if it isn't by its own choice (being checked and all, it gets forwarded everywhere). Although neither version provides as much information as it could, the base version from netty at least makes a compelling case that the author was sure the file vanished (for a moment, or for however long), and there's only two lines to double check that assertion. If I was returning to this class after a long time tracking down such an error, I highly doubt I would immediately suspect the now innocuously placed `setDateAndCacheHeaders` from being any more risky than its neighbors.

tea-dragon:
I made some notes on the code I am most familiar with, but from what I saw in other bits, there are some very reasonable changes. Certainly that `AssetsResource` class was definitely neglecting that `FileInputStream`. I'll see if I can get the respective authors to review the other changes, but you might end up having to make do with me again :)

Once again, we really appreciate the contribution. I hope my comments don't come off the wrong way.

georgekankava:
Hi Ian,

Thank you for the PRs review and sorry that it took some time for you to
add comments.
With this PR you can just let me know which portions should be reverted and
I will leave only necessary ones.
I am looking forward to create some more PRs, maybe more simple ones for
this time.

Regards,
George

PR #199, opened 2016-02-11
squid:S1860 - Synchronization should not be based on Strings or boxed primitives

georgekankava:
squid:S1860 - Synchronization should not be based on Strings or boxed primitives.
https://dev.eclipse.org/sonar/coding_rules#q=squid%3AS1860
George Kankava

### DIFF
--- a/hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterRandom.java
+++ b/hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterRandom.java
@@ -39,6 +39,7 @@ import com.addthis.codec.annotations.FieldConfig;
  */
 public class ValueFilterRandom extends AbstractValueFilter {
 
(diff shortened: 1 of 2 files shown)


tea-dragon:
the change to ValueFilterRandom seems safe enough, but it seems like you have replaced instance level locking with class (static) level locking. However, in FactoryInputStream, there are several calls to eg. `key.wait()` and `key.notifyAll()` which are not okay with the removal of the synchronization on them on the key object. see eg.: https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#wait().

Those are both pretty bizarre classes though. I think i'd end up just gutting most of them myself. As an aside, I generally prefer to go with explicit locks over synchronization blocks. This isn't widely accepted as the best practice or anything, but in my experience, synchronized keywords on methods, blocks, "this", and in static methods tend to lead to all kinds of shenanigans.

PR #200, opened 2016-02-12
squid:S2259 - Null pointers should not be dereferenced

georgekankava:
squid:S2259 - Null pointers should not be dereferenced.
https://dev.eclipse.org/sonar/coding_rules#q=squid%3AS2259
George Kankava

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/job/JobExpand.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/JobExpand.java
@@ -134,8 +134,12 @@ public class JobExpand {
         } catch (InstantiationException | IllegalAccessException e) {
             log.warn("Class '" + clazz + "' registered for '" + macroName + "' cannot be initialized: " + e, e);
         } catch (ClassCastException e) {
-            log.warn("Class '" + clazz + "' registered for '" + macroName + "' is not JobConfigExpander but '" + o.getClass() + "'");
+            log.warn("Class '" + clazz + "' registered for '" + macroName + "' is not JobConfigExpander but '" + getObjectClassIfNotNull(o) + "'");
         }
+    }
+
+     private static String getObjectClassIfNotNull(Object o) {
+        return o != null ? o.getClass().toString() : null;
     }
 
     private static String macroTemplateParamsHelper(String input, final HashMap<String, String> map)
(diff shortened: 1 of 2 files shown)


hydra-main/src/main/java/com/addthis/hydra/job/JobExpand.java
```diff
@@ -134,10 +134,14 @@ static void registerExpander(String macroName, Class<? extends JobConfigExpander
         } catch (InstantiationException | IllegalAccessException e) {
             log.warn("Class '" + clazz + "' registered for '" + macroName + "' cannot be initialized: " + e, e);
         } catch (ClassCastException e) {
-            log.warn("Class '" + clazz + "' registered for '" + macroName + "' is not JobConfigExpander but '" + o.getClass() + "'");
+            log.warn("Class '" + clazz + "' registered for '" + macroName + "' is not JobConfigExpander but '" + getObjectClassIfNotNull(o) + "'");
```
yuesong: o is guaranteed to be not null, so the extra guard is unnecessary here.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/JobTaskItem.java
```diff
@@ -33,7 +33,7 @@ public JobTask getTask() {
 
     @Override
     public boolean equals(Object o) {
-        if (o.getClass() != getClass()) {
+        if (o == null || o.getClass() != getClass()) {
```
tea-dragon: A fun alternative here is a humble instanceof call. It'll return false for both nulls and guys two lines away from a class cast exception.

One time I checked the jvm bytecode specifically to spite an old professor and at the time at least, it was also several instructions more efficient.

hydra-main/src/main/java/com/addthis/hydra/job/JobExpand.java
tea-dragon: Maybe it would be better to split the try block into two. A bit more verbose, but it would be clearer which catch blocks are guarding which line. Something something clarity is great.

It's been an awfully long time, so I suggest closing this PR (with gratitude), and then fixing whatever you want from a cleaner commit.

PR #201, opened 2016-03-11
Spawn task log proxy

mako-taco:
ea09765
Proxies requests for minion tasks through the main spawn server, instead of having the client hit individual minions for logs. This is a step towards allowing spawn completely over https.

1b4ab3d
I also changed the responses spawn sends back from server exceptions. The endpoints should return JSON, even when they error... There is one endpoint, `/expand`, which does an octet stream, but I don't think this is any less correct than it was before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ test.http.get/
 .project
 .settings/
 .idea/
+hydra-auth/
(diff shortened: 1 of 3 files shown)


mako-taco:
redoing a little

PR #202, opened 2016-03-11
Spawn task log proxy

mako-taco:
a504f8b - server errors should be sent as json if the endpoint is supposed to return json
2a52092 - spawn client no longer requests minions themselves for task logs, and instead proxies the request through the main spawn server. This is a step towards having spawn work entirely over HTTPS


hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -1279,7 +1281,26 @@ private static Response buildServerError(Exception exception) {
         if (message == null) {
             message = exception.toString();
         }
-        return Response.serverError().entity(message).build();
+
+        String stack = "";
+        for (StackTraceElement s : exception.getStackTrace()) {
+            stack += s + "\n";
+        }
+
```
yuesong: Use guava Throwables:

`String stack = Throwables.getStackTraceAsString(exception);`

abramsm: Yuesong's approach is best.  Very minor comment on building the string though.  Best practice is to use a StringBuilder rather than building then using String +=.  Java strings are immutable so each concatenation creates a new String.  JVM has reduced penalty for this significantly recently but still beneficial to use StringBuilder approach.

hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -1279,7 +1281,26 @@ private static Response buildServerError(Exception exception) {
...
+        String response;
+
+        try {
+            body.put("error", "A java exception was thrown.");
+            body.put("message", message);
+            body.put("stack", stack);
+            response = body.toString();
+        } catch(JSONException ex) {
+            log.error("[buildServerError]", "failed to construct a json response");
+            response = "{\"error\": \"Failed to serialize java exception. Please see spawn logs for more details.\"}";
```
abramsm: Leaving out the actual exception may lead to confusion later.  The message with the exception would be helpful to know for debugging purposes.

hydra-main/web/spawn2/js/modules/task.log.js
```diff
@@ -29,8 +29,10 @@ function(
...
         },
         url:function(){
-            var url= "http://"+this.host+":"+this.port+"/job.log?out="+(this.stdout?'1':'0')+
-            "&id="+this.jobUuid+"&lines="+this.lines+"&node="+this.node+"&runsAgo="+this.runsAgo;
+            var url= "/job/" + this.jobUuid + "/log?out="+ (this.stdout ? "1" : "0") +
+            "&lines=" + this.lines + "&node=" + this.node + "&runsAgo=" + this.runsAgo +
+            "&minion=" + this.host + "&port=" + this.port;
+
             if(!_.isUndefined(this.offset)){
                 url+="&offset="+this.offset;
```
yuesong: Is the `offset` parameter handled by the minion job.log endpoint? It's not handled by the new spawn endpoint.

hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -862,6 +868,61 @@ public Response getRevisions(@QueryParam("id") String id,
...
+                CloseableHttpClient httpclient = HttpClients.createDefault();
+                HttpGet httpGet = new HttpGet(url);
+                CloseableHttpResponse response = httpclient.execute(httpGet);
+
+                try {
+                    HttpEntity entity = response.getEntity();
+                    String responseEncoding = entity.getContentEncoding() != null ?
+                            entity.getContentEncoding().getValue() :
+                            null;
+                    String encoding = responseEncoding == null ? "UTF-8" : responseEncoding;
```
yuesong: Why not simplify to:
```
String encoding = entity.getContentEncoding() != null ? entity.getContentEncoding().getValue() : "UTF-8";
```

hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -862,6 +868,61 @@ public Response getRevisions(@QueryParam("id") String id,
...
+                return Response.status(Response.Status.BAD_REQUEST).entity(body.toString()).build();
+            } else if (node == null) {
+                body.put("error", "Missing required query parameter 'node'");
+                return Response.status(Response.Status.BAD_REQUEST).entity(body.toString()).build();
+            } else if (port == null) {
+                body.put("error", "Missing required query parameter 'node'");
+                return Response.status(Response.Status.BAD_REQUEST).entity(body.toString()).build();
+            } else {
+                String url = String.format("http://%s:%s/job.log?id=%s&node=%s&runsAgo=%d&lines=%d&out=%d",
+                        minion, port, jobID, node, runsAgo, lines, out);
```
yuesong: Personally I'd use a more expressive API, such as the `URIBuilder` and its `setParameter` method, rather than building the url manually.

hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -862,6 +868,61 @@ public Response getRevisions(@QueryParam("id") String id,
...
+        boolean error = false;
+        try {
+            if (minion == null) {
+                body.put("error", "Missing required query parameter 'minion'");
+                return Response.status(Response.Status.BAD_REQUEST).entity(body.toString()).build();
+            } else if (node == null) {
+                body.put("error", "Missing required query parameter 'node'");
+                return Response.status(Response.Status.BAD_REQUEST).entity(body.toString()).build();
+            } else if (port == null) {
+                body.put("error", "Missing required query parameter 'node'");
```
yuesong: It should be `'port'` in error message, not `'node'`.

hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -890,21 +893,25 @@ public Response getJobTaskLog(@PathParam("jobID") String jobID,
...
+                URI uri = UriBuilder.fromUri("http://" + minion + ":" + port + "/job.log")
+                        .queryParam("id", jobID)
+                        .queryParam("node", node)
+                        .queryParam("runsAgo", runsAgo)
+                        .queryParam("lines", lines)
+                        .queryParam("out", out)
+                        .queryParam("offset", offset)
+                        .build();
+
                 CloseableHttpClient httpclient = HttpClients.createDefault();
```
yuesong: You should reuse the same instance of `HttpClient`, not creating one for each request. See:

http://hc.apache.org/httpcomponents-client-4.5.x/tutorial/html/fundamentals.html#d5e206

mako-taco:
The only ways I know of to close the `HttpClient` before it's deallocated involve `finalize` and `Runtime.addShutdownHook`, which seem not to good.  Is there some jetty specific thing I am missing?

yuesong:
Many jetty components, including the server itself, implement the `LifeCycle` interface, to which you can add a `LifeCycle.Listener` to get notified of life cycle events such as starting, stopping, failure etc. You can add a `destroy` or `shutdown` method to `JobsResource` that closes the httpclient. Then in `SpawnService`'s constructor where the jetty server is instantiated, add a custom life cycle listener to shutdown the resource when server is stopped (use `AbstractLifeCycleListener`).

PR #203, opened 2016-03-16
[Spawn] expanded configuration cache

mako-taco:
Caches expanded configurations so that future work can be done (searching expanded configs) without a bunch of duplicate work by the cpu. Has the added benefit of possibly maybe sometimes speeding up viewing the expanded configs through the UI.


hydra-api/src/main/java/com/addthis/hydra/job/JobParameter.java
```diff
@@ -67,8 +67,30 @@ public void setDefaultValue(String defaultValue) {
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (o.getClass() != JobParameter.class) {
+            return false;
+        }
+
+        JobParameter other = (JobParameter) o;
+
+        boolean nameEquals = (name != null && other.name != null) ? name.equals(other.name) : name == other.name;
```
mspiegel: Use Objects.equals() it will perform the null testing.

mako-taco:
I'll add unit tests if I can get some confirmation that this direction is OK

hydra-api/src/main/java/com/addthis/hydra/job/JobParameter.java
```diff
@@ -74,17 +76,15 @@ public String getParamString() {
 
     @Override
     public boolean equals(Object o) {
-        if (o.getClass() != JobParameter.class) {
+        if (o.getClass() != JobParameter.class || o == null) {
```
yuesong: Null check should be the first to be of any use.

yuesong:
I think the general direction is OK. Some comments:

The cache is added to speed up job expansion, especially when we want to grep all jobs. For the cache to be most effective in that case, it should have most if not all jobs in it. We have about 1000 jobs in each cluster, so we want to make sure memory footprint is reasonable. Our dx-skyfall job is about 60KB when expanded. It should be larger than average. So 1000 jobs are less than 100MB which should be OK. Leaving the default in code at 100 is OK, just make sure the hydra.yaml has the appropriate values.

The cache is invalidated at the right places, so we could use a much larger expiry time setting.

I think we should use job ID as the cache key, not the new class. This means only existing jobs are cache, which is fine. I'm not sure about all the correctness of the equals and hashcode methods added to just support the new key class. Regardless the complexity is not worth the benefit.

mako-taco:
> We should use job ID as the cache key, not the new class

Unfortunately, the cache cannot be used by the UI then, since an unsaved configuration would have the same job ID as the saved one. We currently post the entirety of the config to the server to expand, in this case. It's probably not worth the benefit to use the cache in this case.

> So 1000 jobs are less than 100MB which should be OK. 

There's some other guava cache builder thing called `weigher` which lets you specify a maximum cache based on an arbitrary attribute of your keys/values (I can just do bytes, and then we can limit the cache to XMB)

yuesong:
> Unfortunately, the cache cannot be used by the UI then, since an unsaved configuration would have the same job ID as the saved one. We currently post the entirety of the config to the server to expand, in this case. It's probably not worth the benefit to use the cache in this case.

I don't think that's not true. You just need to move the expansion function out of the cache loader. When it's a call to expand an existing job, you do it via the cache. If it's a new job, you just call the function directly to get the expanded config, no?

yuesong:
> Unfortunately, the cache cannot be used by the UI then, since an unsaved configuration would have the same job ID as the saved one. We currently post the entirety of the config to the server to expand, in this case. It's probably not worth the benefit to use the cache in this case.

You mean when someone's editing a job config and want to see the expand config before saving? The caching doesn't provide any benefit in that case anyway, right?

mako-taco:
if someone opens up a job, and clicks 'expanded config', they will no longer be able to use the cached values (unless we compare the config/params they post to the server with the existing jobs config/params).  It's overly complicated for little gain.  

I have changed it so viewing the expanded the config through the UI completely bypasses the cache, and updated the cache to use weight instead of size.

mako-taco:
Replaced by #206

PR #207, opened 2016-03-22
[spawn-web] modernize web dev environment

mako-taco:
Creates a more modern development environment for web UI work:
 - moves source assets into a new `hydra-web/` folder
 - **builds assets to the existing `hydra-main/web` folders** using webpack
 - **build step is now required to update web-ui**
 - enables newer JS language features (es6, es7)

And as a cheap way to fix a weird AMD/webpack incompatability w.r.t. requesting
data from the server:
 - streaming processing of the initial `/setup/update` response with Oboe


cburroughs:
Please update `NOTICE` if these projects are no longer being vendored.

mako-taco:
Nothing was removed, things were just put in different places. A lot of the 'vendor' things are now listed in `package.json`.

cburroughs:
If we are not including the files verbatim, distributed them, or otherwise doing something that would trigger an attribution requirement, we don't need to put them in `NOTICE`.  Stuff that a build tool downloads or is required at runtime (like the transitive graph of all maven dependencies) does not need to be there. Keeping of track of dependencies in a flat there is a pain, so making it close to the minimal required set eases maintenance.  From the 20 lines added and 14k removed I think this commit is removing the verbatim copies from source control, and using a modern build tool to assemble the web dependencies.

PR #209, opened 2016-04-20
[spawn] new impl of spawn data store, caches results locally

mako-taco:
Preliminary work to move the caching layer (previously at the level of expanded configurations) to the spawn data store.  Using the `MysqlDataStore` without a caching layer has proven to be a big bottleneck when reading attempting to do a full search of all job configs.

A future commit will remove the expanded config cache, and rework how the job config search method works.


mako-taco:
now a part of #211

hydra-main/src/main/java/com/addthis/hydra/job/store/CachedSpawnDataStore.java
```diff
@@ -0,0 +1,128 @@
+package com.addthis.hydra.job.store;
```
tea-dragon: missing license header

PR #213, opened 2016-07-19
workaround for errored jobs blocking fs-ok failure queue

stevedonnelly:
We recently noticed that the fs-ok minion failure queue was being blocked by errored jobs on the hosts being failed.

After running through the debugger and review the code, this appears to be caused by the fs-ok failure reusing logic for normal rebalancing assignments which refuses to move errored tasks.  Since fs-ok failure is supposed to behave similar to normal rebalancing - moving tasks in the background when the cluster has idle task slots - I left this behavior in place.

However, the previous fs-ok failure would only complete the failure once all tasks had been moved off using this rebalancing.  I have changed this condition to fallback to fs-dead failure once only problematic tasks (errored/degraded/etc) remain.

I've tested the change on the test cluster, and it first moved the normal/idle jobs as expected, then performed the typical fs-dead failure by reverting then moving the remaining errored tasks.

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/job/HostFailWorker.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/HostFailWorker.java
@@ -253,6 +253,11 @@ public class HostFailWorker {
                     return;
                 }
                 List<JobTaskMoveAssignment> assignments = spawn.getSpawnBalancer().pushTasksOffDiskForFilesystemOkayFailure(host, tasksToMove);
+                // no re-assignments available for this host, move it to the end of the fs-ok queue
+                if(assignments.isEmpty() && failState == FailState.FAILING_FS_OKAY) {
+                    hostFailState.removeHost(failedHostUuid);
+                    hostFailState.putHost(failedHostUuid, FailState.FAILING_FS_OKAY);
+                }
                 // Use available task slots to push tasks off the host in question. Not all of these assignments will necessarily be moved.
                 spawn.executeReallocationAssignments(assignments, !diskFull && obeyTaskSlots.get());
                 if (failState == FailState.FAILING_FS_OKAY && assignments.isEmpty() && host.countTotalLive() == 0) {
--- a/hydra-main/src/main/java/com/addthis/hydra/job/HostFailWorker.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/HostFailWorker.java
@@ -255,7 +255,7 @@ public class HostFailWorker {
                 List<JobTaskMoveAssignment> assignments = spawn.getSpawnBalancer().pushTasksOffDiskForFilesystemOkayFailure(host, tasksToMove);
                 // Use available task slots to push tasks off the host in question. Not all of these assignments will necessarily be moved.
                 spawn.executeReallocationAssignments(assignments, !diskFull && obeyTaskSlots.get());
-                if (failState == FailState.FAILING_FS_OKAY && assignments.isEmpty()) {
+                if (failState == FailState.FAILING_FS_OKAY && assignments.isEmpty() && host.countTotalLive() == 0) {
                     // Found no tasks on the failed host, so fail it for real.
                     markHostDead(failedHostUuid);
                     spawn.getSpawnBalancer().fixTasksForFailedHost(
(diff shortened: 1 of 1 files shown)


tedpearson:
I have concerns about errored tasks having hosts removed out from under them without a chance to rebalance.

abramsm:
I think this PR is trying to address two issues:

1. fs-ok minion failure hangs when there are errored jobs on the minion
2. hanging/slow fs-ok blocks other fs-oks from completing

For the second issue we could use some sort of time out or error signaling to abort the fs-ok for trouble minions while still allowing other queued failures to proceed.

Ted brings up a valid concern.  Let's imagine a job has an r-factor of 1 and the minion getting fs-ok failed has one of the tasks for that job and it is in error state.  This PR would result in data loss in that case.  Could we add some logic that says something like, "it is OK to fs-dead a minion that has errored jobs IFF the errored tasks on that node have at least one non-errored copy somewhere else in the cluster"?

stevedonnelly:
I discussed Ted's concern with him yesterday and (I think) he was suggesting that fs-dead failing a minion with errored jobs normally results in data loss.  However, I remembered things differently and have re-read the code just to make sure - when fs-dead failing a minion, for each task, any of its replicas is promoted to primary, then a new replica is added to replace the failed minion.  Once this replication is complete, the job state changes back to idle and since the gold links were never touched during the errored run, this effectively reverts all copies of the task to the last successful run.

What I had not considered is that for tasks with only a single copy on the fs-dead host, spawn doesn't fallback to trying to copy from the fs-dead host when no other good replicas are available.  So for errored jobs with 0 replicas, my suggested change would in fact result in unnecessary data loss.

Several possible solutions here:
- the rebalancing logic used by fs-ok could be changed to allow moving of errored jobs (similar to the way fs-dead failure implicitly reverts to last successful run)
- fs-ok failure could be changed to move to the next host in the queue when the current host has no remaining movable tasks (i.e. only errored remain); slightly more complicated change, and would require human intervention to eventually revert/enable the errored tasks that are stopping the failures from completing
- fs-dead failure could be changed to *try* to read from the fs-dead minion when it doesn't have any replicas for the task; could be risky in that it's hard to predict the behavior of fs-dead machines, but it could also prevent data-loss in some hardware failure cases in addition to this fs-ok case

Currently leaning towards the first solution since it appears to be the simplest and would allow our background rebalancing to run on more jobs.  There may be some gotchas that stopped us from doing the implicit reverts in the first place, so I'll probably check with al first.

tedpearson:
I like option 1, however I have concerns about the idea of "implicitly reverting to the last successful run". That seems pretty bad to me. I can see rebalances randomly reverting jobs before people get a chance to look at errors.

 Is there any easy way to allow moving of errored jobs while keeping error? I have suspicions that the state machine that I am guessing exists would become super complicated from this sort of behavior?

stevedonnelly:
While likely possible, retaining errored state after reverting would make both the change and the spawn code in total significantly more complicated.  The rebalance code would need to make sure it replicates from whichever minion has the errored live directory, and then the errored/idle state would need to be passed along until rebalance completion so the correct state could be restored at the end.

Given that auto-revert is generally desirable, and that retaining errored state would really only help debugging non-reproducible job errors (which is quite rare in practice), I think implicit reverts are probably the most reasonable solution.  Then the change essentially becomes adding errored state to a whitelist of rebalance-able states.

alstaplesmoore:
To summarize comments made in Slack, I have no strong objections, but I ask you to consider these cases:

- Suppose a tasks errors on first run, and therefore has a live directory but no backups or replicas. Should we move that busted directory around and reenable the job???
- Suppose a task has some busted files that are causing IO errors when we try to rsync them. If that is the last task remaining, we will very rapidly make a bunch of copies all over the cluster as the task goes from ERROR => REBALANCE => back to ERROR when the rsync fails. You should probably disallow rebalancing if the task failed due to replication/backup issues -- or be very careful in this case.

stevedonnelly:
Decided to go with the second solution - fs-ok queue skips over hosts with only errored tasks remaining - since this is purely an improvement over current behavior, whereas automatically reverting errored tasks could have unexpected consequences (for debugging, and corner cases al mentioned).

Tested behavior on test cluster, and all fs-ok hosts have their non-errored tasks reassigned.  Remaining errored tasks on the fs-ok queued hosts were reassigned after the tasks were manually reverted or enabled.  Finally, the hosts were each changed to failed state once all previously errored tasks completed moving.

I accidentally pushed the previous approach to master, so then I pushed this latest change on top of it.  I think the change should be non-controversial, but we can discuss here if someone has a better idea or sees a problem my testing didn't catch.

PR #214, opened 2016-09-20
add timeZone to BundleFilterTimeRange and add unit test

satang500:
Ref T65100

### DIFF
--- a/hydra-filters/src/main/java/com/addthis/hydra/data/filter/bundle/BundleFilterTimeRange.java
+++ b/hydra-filters/src/main/java/com/addthis/hydra/data/filter/bundle/BundleFilterTimeRange.java
@@ -114,14 +114,12 @@ public class BundleFilterTimeRange implements BundleFilter {
             tafter = 0;
         }
 
-        if( timeFormat != null && timeZone == null ) {
+        if( timeFormat == null ) {
+            this.format = null;
+        } else if ( timeZone == null ) {
             this.format = DateTimeFormat.forPattern(timeFormat);
-        } else if ( timeFormat != null && timeZone != null ) {
-            this.format = DateTimeFormat.forPattern(timeFormat).withZone(DateTimeZone.forID(timeZone));
-        } else if ( timeFormat == null ) {
-            this.format = null;
         } else {
-            this.format = null;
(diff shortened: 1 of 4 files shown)


hydra-filters/src/main/java/com/addthis/hydra/data/filter/bundle/BundleFilterTimeRange.java
```diff
@@ -77,21 +87,22 @@
 
     private final long              tbefore;
     private final long              tafter;
-    private final DateTimeFormatter format;
```
tedpearson: Can't this still be final?

hydra-filters/src/main/java/com/addthis/hydra/data/filter/bundle/BundleFilterTimeRange.java
```diff
@@ -102,6 +113,14 @@ public BundleFilterTimeRange(@JsonProperty(value = "time", required = true) Auto
         } else {
             tafter = 0;
         }
+
+        if ( timeZone != null ) {           // regardless of timeFormat
```
tedpearson: if timeFormat is null here, what happens?

hydra-filters/src/main/java/com/addthis/hydra/data/filter/bundle/BundleFilterTimeRange.java
```diff
@@ -114,10 +114,12 @@ public BundleFilterTimeRange(@JsonProperty(value = "time", required = true) Auto
...
-        if ( timeZone != null ) {           // regardless of timeFormat
-            this.format = DateTimeFormat.forPattern(timeFormat).withZone(DateTimeZone.forID(timeZone));
-        } else if( timeZone == null && timeFormat != null ) {
+        if( timeFormat != null && timeZone == null ) {
             this.format = DateTimeFormat.forPattern(timeFormat);
+        } else if ( timeFormat != null && timeZone != null ) {
+            this.format = DateTimeFormat.forPattern(timeFormat).withZone(DateTimeZone.forID(timeZone));
+        } else if ( timeFormat == null ) {
+            this.format = null;
         } else {
```
tedpearson: This could be reduced to less if statements.

if tf == null;
else if tz == null;
else;

PR #219, opened 2016-09-22
add timezone to bundle filter time-range

satang500:
Updated code

### DIFF
--- a/hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
+++ b/hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
@@ -57,11 +57,6 @@ public class TestBundleFilterTimeRange {
         assertFalse(filter.filter(bundle));
     }
 
-    @Test(expected=JsonMappingException.class)
-    public void testMinusBeforeBadTimeformat() throws IOException {
-        BundleFilterTimeRange filter = Configs.decodeObject(BundleFilterTimeRange.class, "time:TIME, before:-2016/09/23, timeFormat:YYYY/MM/dd");
-    }
-
     @Test
     public void testBeforeAfterInRange() throws IOException {
         String str = "time:TIME, before:20170101, after:20120101, timeFormat:YYYYMMDD";
@@ -86,6 +81,17 @@ public class TestBundleFilterTimeRange {
             fail("Expected exception: time field is required");
         } catch(JsonMappingException jme) {
             assertThat(jme.getMessage(), containsString("marked as required"));
+        }
+    }
+
+    @Test
+    public void testMinusBeforeBadTimeformat() throws IOException {
+        String str = "time:TIME, before:-2016/09/23, timeFormat:YYYY/MM/dd";
+        try {
+            BundleFilterTimeRange filter = Configs.decodeObject(BundleFilterTimeRange.class, str);
+            fail("Expected exception: wrong fomate for date input");
+        } catch(JsonMappingException jme) {
+            assertThat(jme.getMessage(), containsString("value failed: For input string"));
         }
     }
 
--- a/hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
+++ b/hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
@@ -35,9 +35,6 @@ public class TestBundleFilterTimeRange {
         bundle = Configs.decodeObject(Bundle.class, "TIME:1474545600000");   // 09/22/2016
     }
 
-//"filtered out later than 160923" meant given time is not filtered out due to "filtered out later than 160923" reason. I agree that this expression can be misread. Will change it.
-
-
     @Test
     public void testBeforeNoTimezone() throws IOException {
         BundleFilterTimeRange filter1 = Configs.decodeObject(BundleFilterTimeRange.class, BEFORE_160923);
--- a/hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
+++ b/hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
@@ -87,13 +87,7 @@ public class TestBundleFilterTimeRange {
 
     @Test(expected=JsonMappingException.class)
     public void testNoTimeformat() throws IOException {
-        String str = "time:TIME, before:20170101, after:20120101, defaultExit:true, timeFormat:null, timeZone:EST";
-        BundleFilterTimeRange filter = Configs.decodeObject(BundleFilterTimeRange.class, str);
-    }
-
-    @Test(expected=JsonMappingException.class)
-    public void testNoTimeformatNoTimezone() throws IOException {
-        String str = "time:TIME, before:20170101, after:20120101, defaultExit:true, timeFormat:null, timeZone:null";
+        String str = "time:TIME, before:20170101, after:20120101, defaultExit:true, timeZone:EST";
         BundleFilterTimeRange filter = Configs.decodeObject(BundleFilterTimeRange.class, str);
     }
 }
--- a/hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
+++ b/hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
@@ -13,38 +13,69 @@
  */
 package com.addthis.hydra.data.filter.bundle;
 
+import com.addthis.bundle.core.Bundle;
 import com.addthis.bundle.util.AutoField;
+import com.addthis.codec.config.Configs;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.format.DateTimeFormatter;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.io.IOException;
 import java.text.DecimalFormat;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
+import static org.junit.Assert.*;
 import static org.mockito.Mockito.mock;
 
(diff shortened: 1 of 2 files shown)


hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
```diff
@@ -13,38 +13,69 @@
...
 public class TestBundleFilterTimeRange {
     AutoField timeIn ;
     DateTime utc;
     DecimalFormat df;
+    Bundle bundle;
+
+    static final String before1 = "time:TIME, before:2016-09-23, timeFormat:YYYY-MM-dd";
+    static final String before2 = "time:TIME, before:2015-09-25, timeFormat:YYYY-MM-dd";
+    static final String after1  = "time:TIME,  after:2016-09-21, timeFormat:YYYY-MM-dd";
+    static final String after2  = "time:TIME,  after:2017-09-21, timeFormat:YYYY-MM-dd";
```
yuesong: All constants above should be `private` too. Also the naming convention for constants are all CAPS. The names can be better too. For example, instead of `before1`, consider `BEFORE_160923` - it makes the test methods below easier to read.

hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
```diff
@@ -13,38 +13,69 @@
...
         df = new DecimalFormat("00");
+        bundle = Configs.decodeObject(Bundle.class, "TIME:1474502400000");   // 09/22/2016
+    }
+
+    @Test
+    public void testBeforeNoTimeZone() throws IOException {
+        BundleFilterTimeRange filter1 = Configs.decodeObject(BundleFilterTimeRange.class, before1);
+        BundleFilterTimeRange filter2 = Configs.decodeObject(BundleFilterTimeRange.class, before2);
+        assertTrue(filter1.filter(bundle));
+        assertFalse(filter2.filter(bundle));
```
yuesong: When there are multiple assertions as above, use `assertTrue(String, boolean)` to provide a description of the assertion. Test failure output will be easier to read too.

hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
```diff
@@ -13,38 +13,69 @@
...
     @Test
-    public void testBundleFilterTimeRangeAB() {
+    public void testAfterNoTimeZone() throws IOException {
+        BundleFilterTimeRange filter1 = Configs.decodeObject(BundleFilterTimeRange.class, after1);
+        BundleFilterTimeRange filter2 = Configs.decodeObject(BundleFilterTimeRange.class, after2);
+        assertTrue(filter1.filter(bundle));
+        assertFalse(filter2.filter(bundle));
+    }
+
+    @Test(expected= Exception.class)
```
yuesong: I assume it is a subclass of Exception expected here? If so, you should use the exact type, not the catchall `Exception`.

hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
```diff
@@ -13,121 +13,107 @@
...
-        DateTimeFormatter format = bftr.getTimeZoneFormat();
-
-        String time = format.parseDateTime(input).getYear() + "" +
-                df.format(format.parseDateTime(input).getMonthOfYear()) + "" +
-                df.format(format.parseDateTime(input).getDayOfMonth()) + "" +
-                df.format(format.parseDateTime(input).getHourOfDay()) + "" +
-                df.format(format.parseDateTime(input).getMinuteOfHour());
-        assertEquals(input, time.substring(2));
+    public void testMinusBefore() throws IOException {
+        BundleFilterTimeRange filter = Configs.decodeObject(BundleFilterTimeRange.class, "time:TIME, before:-20160923, timeFormat:YYYYMMdd");
```
yuesong: `before:-20160923, timeFormat:YYYYMMdd` is misleading as the reader may think the value for `before` should match `timeFormat`, while in reality a value starting with a `-` indicates number of days and `timeFormat` is not used to parse it. Something like `before:-10000` would be better.

hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
```diff
@@ -13,121 +13,107 @@
...
+        assertFalse(filter.filter(bundle));
     }
 
     @Test
-    public void testBundleFilterTimeRangeNoTimeFormat() {
-        DateTimeZone tz = DateTimeZone.forID("EST");
-        DateTime ntzntfDateTime = utc.toDateTime(tz);
-        String input = ntzntfDateTime.toString("yyMMddHHmm");
+    public void testNoTime() throws IOException {
+        String str = "time:null, before:20170101, after:20120101,timeFormat:YYYYMMDD";
```
yuesong: `time:null` can be left out.

hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
```diff
@@ -13,121 +13,107 @@
...
+        try {
+            BundleFilterTimeRange filter = Configs.decodeObject(BundleFilterTimeRange.class, str);
+            fail("Expected exception: time field is required");
+        } catch(JsonMappingException jme) {
+            assertThat(jme.getMessage(), containsString("marked as required"));
```
yuesong: Verifying exception message here is OK, but the main verification is that an exception is thrown. Using `expected=JsonMappingException.class` is good enough and easier to read.

hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
```diff
@@ -13,121 +13,107 @@
...
-                df.format(format.parseDateTime(input).getMonthOfYear()) + "" +
-                df.format(format.parseDateTime(input).getDayOfMonth()) + "" +
-                df.format(format.parseDateTime(input).getHourOfDay()) + "" +
-                df.format(format.parseDateTime(input).getMinuteOfHour());
-        assertEquals(input, time.substring(2));
+    public void testBeforeNoTimezone() throws IOException {
+        BundleFilterTimeRange filter1 = Configs.decodeObject(BundleFilterTimeRange.class, BEFORE_160923);
+        BundleFilterTimeRange filter2 = Configs.decodeObject(BundleFilterTimeRange.class, BEFORE_150925);
+        assertTrue("Expected true since filtered out later than 160923", filter1.filter(bundle));
+        assertFalse("Expected false since filtered out later than 150925", filter2.filter(bundle));
```
yuesong: Both messages say "later than" - one of them must be wrong.

hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
```diff
@@ -13,121 +13,107 @@
...
-                df.format(format.parseDateTime(input).getMonthOfYear()) + "" +
-                df.format(format.parseDateTime(input).getDayOfMonth()) + "" +
-                df.format(format.parseDateTime(input).getHourOfDay()) + "" +
-                df.format(format.parseDateTime(input).getMinuteOfHour());
-        assertEquals(input, time.substring(2));
+    public void testAfterNoTimezone() throws IOException {
+        BundleFilterTimeRange filter1 = Configs.decodeObject(BundleFilterTimeRange.class, AFTER_160921);
+        BundleFilterTimeRange filter2 = Configs.decodeObject(BundleFilterTimeRange.class, AFTER_170921);
+        assertTrue("Expected true since filtered out earlier than 160921", filter1.filter(bundle));
+        assertFalse("Expected false since filtered out earlier than 170921", filter2.filter(bundle));
```
yuesong: Same here regarding the messages.

hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
```diff
@@ -13,121 +13,87 @@
...
-        String input = ntzntfDateTime.toString("yyMMddHHmm");
+    @Test(expected=JsonMappingException.class)
+    public void testMinusBeforeBadTimeformat() throws IOException {
+        String str = "time:TIME, before:-1000/00";
+        BundleFilterTimeRange filter = Configs.decodeObject(BundleFilterTimeRange.class, str);
+    }
+
+    @Test(expected=JsonMappingException.class)
+    public void testNoTimeformat() throws IOException {
+        String str = "time:TIME, before:20170101, after:20120101, defaultExit:true, timeFormat:null, timeZone:EST";
```
yuesong: `timeFormat:null` is unnecessary

hydra-filters/src/test/java/com/addthis/hydra/data/filter/bundle/TestBundleFilterTimeRange.java
```diff
@@ -13,121 +13,87 @@
...
+        BundleFilterTimeRange filter = Configs.decodeObject(BundleFilterTimeRange.class, str);
+    }
 
-        BundleFilterTimeRange bftr = new BundleFilterTimeRange( timeIn, "121201", "130101", true, null, null);
-        DateTimeFormatter format = bftr.getTimeZoneFormat();
-        assertNull(format);
+    @Test(expected=JsonMappingException.class)
+    public void testNoTimeformatNoTimezone() throws IOException {
+        String str = "time:TIME, before:20170101, after:20120101, defaultExit:true, timeFormat:null, timeZone:null";
```
yuesong: `timeFormat:null` and `timeZone:null` are unnecessary. Also isn't this test redundant given the previous one?

PR #220, opened 2016-10-15
Add support for job alert webhook events

jhorwit2:
todo still / questions:
- [x] - test ui changes
- [ ] - update alert docs (do they exist?)
- [x] - cleanup json event object to be less emaily.


yuesong:
While I'm reviewing the code, @jhorwit2 have you tested your change in local stack?

jhorwit2:
@yuesong working.


output from webhook received by hydra

```
josh@josh-VirtualBox:~$ ./test.py 
Listening on localhost:8080

----- Request Start ----->

/
Content-Length: 427
Content-Type: application/json
Host: localhost:8080
Connection: Keep-Alive
User-Agent: Apache-HttpClient/4.4.1 (Java/1.8.0_91)
Accept-Encoding: gzip,deflate

{"alert_description":"","alert_link":"http://localhost:5052/spawn2/index.html#alerts/fbbdb698-84cd-4f35-adf1-e7dcd42455e2","alert_type":"Task is in Error","alert_reason":"[TRIGGER]","jobs_in_error":[{"id":"7ebb02e4-c12f-4b21-9faa-073b8bf07145","description":"(no title)","error":"Task 0 -> 127","cluster_head":"localhost","job_state":"ERROR","start_time":1476725405867,"end_time":1476729435312,"node_count":1,"error_count":0}]}
<----- Request End -----

127.0.0.1 - - [17/Oct/2016 15:56:35] "POST / HTTP/1.1" 200 -
```

hydra-main/src/main/java/com/addthis/hydra/job/alert/JobAlertRunner.java
```diff
@@ -241,15 +254,96 @@ private static String format(Long time) {
...
+
+    private void sendWebhookAlert(AbstractJobAlert jobAlert,
+                                  String alertLink,
+                                  String reason,
+                                  Map<String, String> errors) {
+
+        try {
+            byte[] body = objectMapper.writeValueAsBytes(getWebhookObject(spawn, jobAlert, alertLink, reason, errors));
+            HttpResponse response = HttpUtil.httpPost(jobAlert.webhookURL, ContentType.APPLICATION_JSON.getMimeType(), body, 5_000);
+            if (response.getStatus() / 100 != 2) {
```
jhorwit2: redirect won't return a 3xx status code when enabled in an apache client. it will return the status code of the request after the redirect.

yuesong: sure. that's what i meant by checking `>=400` if the post method does auto redirect

yuesong:
What's your thought on multiple webhook urls support?

jhorwit2:
Do the post hooks in job settings support multiple urls? If not, then I'd say no. I also can't think of a use case where we would need multiple. It's not like you can use the webhook to post to PD or other apps because it's not the correct format, so something will have to consume it then send it out to appropriate places.

hydra-main/src/main/java/com/addthis/hydra/job/alert/JobAlertRunner.java
jhorwit2: It only returns a redirect status code when redirects are disabled and it won't follow the redirect.

jhorwit2: If the post doesn't redirect then that means the webhook event failed and we should log it as an error, because the redirect never happened. Otherwise, we are swallowing the failed redirect status code.

jhorwit2:
thanks @yuesong !

PR #221, opened 2016-10-28
Fix unbox NPE issue from Long to long in job runner

jhorwit2:


tedpearson:
Looks good to me, merging

hydra-main/src/main/java/com/addthis/hydra/job/alert/JobAlertRunner.java
```diff
@@ -299,8 +299,14 @@ static AlertWebhookRequest getWebhookObject(Spawn spawn,
             if (job != null) {
                 jobError.setJobState(job.getState());
                 jobError.setDescription(job.getDescription());
-                jobError.setStartTime(job.getStartTime());
-                jobError.setEndTime(job.getEndTime());
+
+                if (job.getStartTime() != null) {
```
tea-dragon: Are you quite sure this is all thread safe? I don't see any use of the job lock or other guards, and it seems quite plausible to me that the start or end time could become null again between calls (anyone or anything that eg. reschedules/kicks the job will immediately null them out).

A common pattern for this particular race condition is to first store the uncertain variable into a friendly, unchanging local variable. In fact my intellij inspections should have issued grave warnings to that effect.

Anyway, I'd take a closer look at the other ways jobs are being used here as well...

PR #222, opened 2016-11-10
Added include/exclude to json outputstream just like channel

jhorwit2:


stevedonnelly:
have you tested this when both include and exclude are null?  based on a quick read of FIlteredBundle it looks like it might NPE: https://github.com/addthis/hydra/blob/master/hydra-task/src/main/java/com/addthis/hydra/task/output/FilteredFormat.java#L55

jhorwit2:
I copy pasted this logic directly from https://github.com/addthis/hydra/blob/ffd708a1078f2bcb24cc873d4f12ef18f9d410b4/hydra-task/src/main/java/com/addthis/hydra/task/output/OutputStreamChannel.java

I'm very surprised no one has hit that condition yet. I can add an if not null on exclude?

jhorwit2:
Actually this won't NPE because the class will only make the filter if ` if (include != null || exclude != null) {`

tea-dragon:
Is there some reason filtered data output did not suffice?

PR #225, opened 2016-12-21
Donnelly/blocked query metrics

stevedonnelly:
Ive already tested this on the test cluster and it is working as intended.

I tried to reduce the minimize the amount of code added, rather than optimize it more than necessary.  Namely, the blockedTaskStats method is called twice - once for the number of blocked stats, once for the maximum task wait time.  These gauge metrics are only updated once a minute, only iterate over ~10 queries typically, and are not blocking any important threads.  Thus, I thought the trade off was worth while.


tea-dragon:
🏎

PR #226, opened 2017-01-09
addresses replicating-kicking race condition in minion

stevedonnelly:
Currently, failing a minion triggers a large number of replicating tasks, many which become errored due to failed kicks.
I believe this is due to a race condition where the kick message is sent to spawn while the task is idle, then a minion
failure causes spawn to send a task-replicate message to the same minion before the kick message in processesd, the minion
handles the replicate message, finally the minion tries to handle the kick message but fails because the task isnt idle.
This change essentially demotes the non-idle failure to a warning, similar to the minions response when it tries to kick
a task it doesnt have run-capacity for.

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/minion/JobTask.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/minion/JobTask.java
@@ -25,6 +25,7 @@ import java.util.List;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.function.Function;
 
 import com.addthis.basis.util.LessBytes;
 import com.addthis.basis.util.LessFiles;
@@ -752,9 +753,13 @@ public class JobTask implements Codable {
         return backups.get(offset);
     }
 
-    private void require(boolean test, String msg) throws ExecException {
+    private void require(boolean test, String msg) throws Exception {
+        require(test, msg, ExecException::new);
+    }
+
+    private void require(boolean test, String msg, Function<String,Exception> exceptionType) throws Exception {
         if (!test) {
-            throw new ExecException(msg);
+            throw exceptionType.apply(msg);
         }
     }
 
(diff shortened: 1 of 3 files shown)


stevedonnelly:
Just so it's on the record, there's a problem with this branch in that went spawn receives the cant-begin-task message, it tries to set the task state to idle (assuming it was previously in allocated state).

So, this shouldnt be merged until I push a fix.  Still trying to decide how I want to handle this, probably send back a different message type when the task is replicating/non-idle.

hydra-main/src/main/java/com/addthis/hydra/minion/JobTask.java
```diff
@@ -789,7 +794,7 @@ public void exec(@Nonnull CommandTaskKick kickMessage, boolean execute) throws E
             String jobId = kickMessage.getJobUuid();
             int jobNode = kickMessage.getJobKey().getNodeNumber();
             log.debug("[task.exec] {}", kickMessage.getJobKey());
-            require(testTaskIdle(), "task is not idle");
+            require(testTaskIdle(), "task is not idle", ExecStateException::new);
```
yuesong: It'd be better to include the actual non-idle state, perhaps job id and task id too, in the exception message

stevedonnelly:
I implemented your suggestion of adding the current task state to the exception message.  The job and task ids would already be logged in Minion where it catches the exception, but not in the exception message itself.

As for the sending the ignored kick message back to spawn, I dont think there's any benefit operationally, and there is some cost in terms of code size (new message type, new case in the if/else chain of message type handlers).

The yammer metrics are already centralized, and I think it would almost always be easier to check for the ignored-kick problem happening via metrics rather than searching the spawn logs.  Furthermore, once you do see that a kick is ignored, you will probably want to find the specific minion that had the problem and go to its logs from that time period - I don't think there's any avoiding that.

tedpearson:
While it may be easier to check metrics than spawn log, convention is still spawn log. I never look at hydra metrics.

PR #227, opened 2017-02-24
Disabling alert when query system is down ref T67375

satang500:


yuesong:
When mqmaster is down, do we get something more specific than IOException (like some kind of connection exception)? I think the error handling code for transient SocketTimeoutException can be reused here. IOException is too generic, but I'm ok with handling a more specific exception in the same way as SocketTimeoutException. If we can do that, there's no need for the additional transient field, or the change to JobAlertRunner.

PR #229, opened 2017-03-20
ValueFilterMapValue improvements

stevedonnelly:
- changing key and other inputs to AutoFields (rather than string constants)
    - adding support for getWithDefault, and put (returns a new ValueMap with added kv entry)
    - adding unit test

### DIFF
--- a/hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterMapValue.java
+++ b/hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterMapValue.java
@@ -55,17 +55,17 @@ public class ValueFilterMapValue extends AbstractValueFilterContextual {
 
     @Override
     public ValueObject filterValue(ValueObject value, Bundle context) {
-        if (value == null || value.getObjectType() != ValueObject.TYPE.MAP) {
+        if (value == null || value.getObjectType() != ValueObject.TYPE.MAP || this.key == null) {
             // TODO: log error
             return null;
         }
         ValueMap mapValue = value.asMap();
-        if(this.put != null) {
+        if (this.put != null) {
             ValueMap copy = ValueFactory.createMap();
             copy.putAll(mapValue);
             copy.put(this.key.getString(context).get(), this.put.getValue(context));
             return copy;
-        } else if(this.defaultValue != null) {
+        } else if (this.defaultValue != null) {
             return mapValue.getOrDefault(this.key.getString(context).get(), this.defaultValue.getValue(context));
         } else {
             return mapValue.get(this.key.getString(context).get());
--- a/hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterMapValue.java
+++ b/hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterMapValue.java
@@ -13,36 +13,62 @@
  */
 package com.addthis.hydra.data.filter.value;
 
+import com.addthis.bundle.core.Bundle;
+import com.addthis.bundle.util.AutoField;
+import com.addthis.bundle.util.AutoParam;
+import com.addthis.bundle.value.ValueFactory;
 import com.addthis.bundle.value.ValueMap;
 import com.addthis.bundle.value.ValueObject;
 import com.addthis.codec.annotations.FieldConfig;
 
 /**
- * This {@link AbstractValueFilter ValueFilter} <span class="hydra-summary">returns the value associated with a specific key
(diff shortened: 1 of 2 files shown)


hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterMapValue.java
```diff
@@ -13,36 +13,62 @@
...
         if (value == null || value.getObjectType() != ValueObject.TYPE.MAP) {
             // TODO: log error
             return null;
         }
         ValueMap mapValue = value.asMap();
-        return mapValue.get(key);
+        if(this.put != null) {
+            ValueMap copy = ValueFactory.createMap();
+            copy.putAll(mapValue);
+            copy.put(this.key.getString(context).get(), this.put.getValue(context));
```
jhorwit2: what if key is null. I don't see any tests for this. 

What's the reasoning for shallow copy of the map? This changes the result and possibly impacts people who assumed it would not be a shallow copy.

hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterMapValue.java
```diff
@@ -13,36 +13,62 @@
...
     @Override
-    public ValueObject filterValue(ValueObject value) {
+    public ValueObject filterValue(ValueObject value, Bundle context) {
         if (value == null || value.getObjectType() != ValueObject.TYPE.MAP) {
             // TODO: log error
             return null;
         }
         ValueMap mapValue = value.asMap();
-        return mapValue.get(key);
+        if(this.put != null) {
```
jhorwit2: syntax is messed up. doesn't follow style guides we have

stevedonnelly: what's the style issue specifically?

hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterMapValue.java
stevedonnelly: So, the key cannot be null, but it doesnt seem like there's a way to specified the "required" flag on the @AutoParam annotation.  Maybe I can use both @JsonProperty and @AutoParam, although I didn't find any examples that do something like this.

Reason for shallow copy is that *ValueFilters* are essentially stateless transforms on ValueObjects - you might have been thinking of BundleFilters which are intended to change the underlying bundle.  Specific example:
`{from: input_map, to: output_map, map-value: {key: a, put: 1}}`
would create a new field, `output_map` that would contain the new entry (a, 1), whereas `input_map` would be unchanged.

Since the `put` option to the value filter is entirely new, this behavior shouldn't break any existing uses.

hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterMapValue.java
jhorwit2: spacing. can't have all this anarchy :trollface:

hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterMapValue.java
jhorwit2: What can't the key be null? I don't see anything preventing it. Obviously it's useless if it is null but an IllegalArguementException or something is better than an NPE.

jhorwit2: Actually sorry, being null before would just return null. In this case it would throw an NPE. 

This goes for here with `put` and below with the `get` calls.

hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterMapValue.java
```diff
@@ -13,36 +13,62 @@
...
+            ValueMap copy = ValueFactory.createMap();
+            copy.putAll(mapValue);
+            copy.put(this.key.getString(context).get(), this.put.getValue(context));
+            return copy;
+        } else if(this.defaultValue != null) {
+            return mapValue.getOrDefault(this.key.getString(context).get(), this.defaultValue.getValue(context));
+        } else {
+            return mapValue.get(this.key.getString(context).get());
```
jhorwit2: This would NPE now versus return null like before when key is null.

PR #232, opened 2017-04-06
Introduce rebalance metrics & track effectiveness

tedpearson:
This is a prerequisite for a rework of hydra rebalancing, hopefully coming soon.

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostScore.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostScore.java
@@ -16,13 +16,11 @@ package com.addthis.hydra.job.spawn.balancer;
 final class HostScore {
 
     private final double meanActiveTasks;
-    private final long freeDiskBytes;
     private final double usedDiskPercent;
     private final double overallScore;
 
-    HostScore(double meanActiveTasks, double usedDiskPercent, long diskFreeBytes, double overallScore) {
+    HostScore(double meanActiveTasks, double usedDiskPercent, double overallScore) {
         this.meanActiveTasks = meanActiveTasks;
-        this.freeDiskBytes = diskFreeBytes;
         this.usedDiskPercent = usedDiskPercent;
         this.overallScore = overallScore;
     }
@@ -33,9 +31,5 @@ final class HostScore {
 
     public double getScoreValue(boolean diskSpace) {
         return diskSpace ? usedDiskPercent : meanActiveTasks;
-    }
-
-    public long getFreeDiskBytes() {
-        return freeDiskBytes;
     }
 }
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostScore.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostScore.java
@@ -16,11 +16,13 @@ package com.addthis.hydra.job.spawn.balancer;
 final class HostScore {
 
     private final double meanActiveTasks;
+    private final long diskFreeBytes;
     private final double usedDiskPercent;
     private final double overallScore;
 
-    HostScore(double meanActiveTasks, double usedDiskPercent, double overallScore) {
+    HostScore(double meanActiveTasks, double usedDiskPercent, long diskFreeBytes, double overallScore) {
         this.meanActiveTasks = meanActiveTasks;
+        this.diskFreeBytes = diskFreeBytes;
         this.usedDiskPercent = usedDiskPercent;
         this.overallScore = overallScore;
     }
@@ -31,5 +33,9 @@ final class HostScore {
 
     public double getScoreValue(boolean diskSpace) {
         return diskSpace ? usedDiskPercent : meanActiveTasks;
+    }
+
+    public long getDiskFreeBytes() {
+        return diskFreeBytes;
     }
 }
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostScore.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostScore.java
@@ -16,13 +16,13 @@ package com.addthis.hydra.job.spawn.balancer;
 final class HostScore {
 
     private final double meanActiveTasks;
-    private final long diskFreeBytes;
+    private final long freeDiskBytes;
     private final double usedDiskPercent;
     private final double overallScore;
 
     HostScore(double meanActiveTasks, double usedDiskPercent, long diskFreeBytes, double overallScore) {
         this.meanActiveTasks = meanActiveTasks;
-        this.diskFreeBytes = diskFreeBytes;
+        this.freeDiskBytes = diskFreeBytes;
         this.usedDiskPercent = usedDiskPercent;
         this.overallScore = overallScore;
     }
@@ -35,7 +35,7 @@ final class HostScore {
         return diskSpace ? usedDiskPercent : meanActiveTasks;
     }
 
-    public long getDiskFreeBytes() {
-        return diskFreeBytes;
+    public long getFreeDiskBytes() {
+        return freeDiskBytes;
     }
 }
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
@@ -32,6 +32,7 @@ import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
 
 import java.math.BigInteger;
 
@@ -173,20 +174,21 @@ public class SpawnBalancer implements Codable, AutoCloseable {
     }
 
     private void initMetrics() {
-        Metrics.newGauge(this.getClass(), "minDiskFreeDiff", SpawnBalancer.makeGauge(minDiskFreeDiff));
-        Metrics.newGauge(this.getClass(), "maxDiskFreeDiff", SpawnBalancer.makeGauge(maxDiskFreeDiff));
-        Metrics.newGauge(this.getClass(), "avgDiskFreeDiff", SpawnBalancer.makeGauge(avgDiskFreeDiff));
-        Metrics.newGauge(this.getClass(), "minTaskPercentDiff", SpawnBalancer.makeGauge(minTaskPercentDiff));
-        Metrics.newGauge(this.getClass(), "minTaskPercentDiff", SpawnBalancer.makeGauge(minTaskPercentDiff));
-        Metrics.newGauge(this.getClass(), "minTaskPercentDiff", SpawnBalancer.makeGauge(minTaskPercentDiff));
+        SpawnBalancer.makeGauge("minDiskFreeDiff", () -> minDiskFreeDiff);
+        SpawnBalancer.makeGauge("maxDiskFreeDiff", () -> maxDiskFreeDiff);
+        SpawnBalancer.makeGauge("avgDiskFreeDiff", () -> avgDiskFreeDiff);
+        SpawnBalancer.makeGauge("minTaskPercentDiff", () -> minTaskPercentDiff);
+        SpawnBalancer.makeGauge("minTaskPercentDiff", () -> minTaskPercentDiff);
+        SpawnBalancer.makeGauge("minTaskPercentDiff", () -> minTaskPercentDiff);
     }
 
-    private static <T> Gauge<T> makeGauge(T value) {
-        return new Gauge<T>() {
+    private static <T> void makeGauge(String name, Supplier<T> value) {
+        Gauge<T> gauge = new Gauge<T>() {
             @Override public T value() {
-                return value;
+                return value.get();
             }
         };
+        Metrics.newGauge(SpawnBalancer.class, name, gauge);
     }
 
     /** Loads SpawnBalancerConfig from data store; if no data or failed, returns the default. */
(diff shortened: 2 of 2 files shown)


hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1225,6 +1306,7 @@ private HostScore calculateHostScore(HostState host, double clusterMaxMeanActive
         double score = 0;
         double meanActive = host.getMeanActiveTasks();
         double usedDiskPercent = getUsedDiskPercent(host);
+        long diskFreeBytes = host.getMax().getDisk() - host.getUsed().getDisk();
```
yuesong: Add a new helper method `HostState.getDiskFreeBytes()` to do this.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1205,14 +1246,54 @@ protected void updateAggregateStatistics(Iterable<HostState> hosts) {
...
                 for (HostState host : hosts) {
-                    cachedHostScores.put(host.getHostUuid(),
-                                         calculateHostScore(host, maxMeanActive, maxDiskPercentUsed));
+                    // metrics here
+                    HostScore score = calculateHostScore(host, maxMeanActive, maxDiskPercentUsed);
+                    cachedHostScores.put(host.getHostUuid(), score);
+                    long freeDisk = score.getDiskFreeBytes();
+                    double taskPercent = score.getScoreValue(false);
+                    sumDiskFree.add(BigInteger.valueOf(freeDisk));
+                    sumTaskPercent += taskPercent;
```
yuesong: Not quite sure what taskPercent really is. Do all hosts have the same denominator for this percentage?
 If not, summing and averaging percentage feels a bit strange...

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostScore.java
```diff
@@ -32,4 +34,8 @@ public double getOverallScore() {
     public double getScoreValue(boolean diskSpace) {
```
tedpearson: No. But this is just preliminary work before trying to fix everything wrong with this code.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1205,14 +1246,54 @@ protected void updateAggregateStatistics(Iterable<HostState> hosts) {
                 findActiveJobIDs();
                 double maxMeanActive = -1;
                 double maxDiskPercentUsed = -1;
+                long maxDiskFree = 0;
+                long minDiskFree = Long.MAX_VALUE;
+                BigInteger sumDiskFree = BigInteger.valueOf(0);
```
tedpearson: If you add 200+ longs together, it's possible to overflow a long

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
tedpearson: taskPercent is a value from 0 to 1, indicating how many percent of task slots have been taken up on average since the minion process started.

Currently, we only have 3 tasks per minion. That will change. So just getting the metric directly here doesn't make sense for the future.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1205,14 +1246,54 @@ protected void updateAggregateStatistics(Iterable<HostState> hosts) {
...
+                    // metrics here
+                    HostScore score = calculateHostScore(host, maxMeanActive, maxDiskPercentUsed);
+                    cachedHostScores.put(host.getHostUuid(), score);
+                    long freeDisk = score.getDiskFreeBytes();
+                    double taskPercent = score.getScoreValue(false);
+                    sumDiskFree.add(BigInteger.valueOf(freeDisk));
+                    sumTaskPercent += taskPercent;
+                    if (freeDisk > maxDiskFree) {
+                        maxDiskFree = freeDisk;
+                    }
```
tedpearson: Yes! I knew there had to be a better way, thanks.

tedpearson:
>Consider moving some of the static helper methods getUsedDiskPercent and getAvailableDiskBytes which take a single HostState as parameter to HostState class

Sure, great ideas, but maybe when I actually try to make the code good instead of when I just want some metrics so I can have a baseline before fixing code.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
yuesong: In our specific case, Long.MAX_VALUE = 2^63-1, and 1 petabyte = 2^50, so `long` should be more than enough here. It makes the code a bit easier to read, that's all.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
tedpearson: There's a static method already in this class, will use that for now and move later

tedpearson:
I made the changes I feel are important to make now. Let me know if anything else needs looking at.

yuesong:
Copying my comments made separately:

the `makeGuage` method will make a gauge that return the constant initial value which is 0. Need to use a lambda. You might as well put calls to `Metrics.newGauge` in the `makeGauge` method too.

tedpearson:
I *think* we're good now.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1185,14 +1213,46 @@ protected void updateAggregateStatistics(Iterable<HostState> hosts) {
...
+            double sumTaskPercentDiff = 0;
+            for (HostScore score : cachedHostScores.values()) {
+                long diskDiff = Math.abs(avgDiskFree - score.getFreeDiskBytes());
+                sumDiskFreeDiff += diskDiff;
+                sumTaskPercentDiff += Math.abs(avgTaskPercent - score.getScoreValue(false));
+            }
+            avgDiskFreeDiff = sumDiskFreeDiff / (long) numScores;
+            avgTaskPercentDiff = sumTaskPercentDiff / (double) numScores;
+            minDiskFreeDiff = avgDiskFree - minDiskFree;
+            maxDiskFreeDiff = maxDiskFree - avgDiskFree;
```
jhorwit2: Using this won't be beneficial when you have a heterogenous cluster. You'll want to normalize based on total disk space also.

tedpearson: Why?

jhorwit2: Typically when you are writing scores for a scheduler you want to normalize the variables so your scheduler is fair. 

I.e. 
`(total - free) / total`

The current way you are keeping track of things works fine when your cluster is homogeneous, but it will start to be unfair when nodes of different sizes enter the picture.

tedpearson: This is a metric for tracking rebalancer effiency. It is not a scheduler nor does it affect scheduling.

jhorwit2: For tracking effectiveness, z-score may also be more useful.

jhorwit2: Also, HostScore is used for scheduling whether it's job scheduling or rebalancing, so adding efficiency metrics to that is a tad weird to me. 


The rebalancer, which is a scheduler, should track efficiency based on fair distribution. You want to evenly balance your cluster, right?

yuesong:
Gauge creation looks good to me.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostScore.java
```diff
@@ -16,11 +16,13 @@
...
 
     private final double meanActiveTasks;
+    private final long diskFreeBytes;
     private final double usedDiskPercent;
     private final double overallScore;
 
-    HostScore(double meanActiveTasks, double usedDiskPercent, double overallScore) {
+    HostScore(double meanActiveTasks, double usedDiskPercent, long diskFreeBytes, double overallScore) {
         this.meanActiveTasks = meanActiveTasks;
+        this.diskFreeBytes = diskFreeBytes;
```
tea-dragon: This doesn't seem immediately useful to the purported purpose of calculating the host score. The only use I see appears to still have host states themselves in scope (or very near).

When dealing with spawn, I suggest thinking twice before adding a line of code, and thinking five more times before adding fields.

It isn't too late to add it later if it becomes part pf the actual score or external function of the class.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
tea-dragon: No changes have been made to the core logic of HostScore. It uses heuristics developed by Al, but they are based on the average percentage use of task slots and average percentage use of disk space.

There are also more than a few configuration parameters that can be adjusted, and something like excess disk space has myriad solutions. Many of which are better than potentially overloading a minion's other resources.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostScore.java
tedpearson: @tea-dragon - you are right. I will remove this and just use the host states that are in scope. Thanks for the feedback!

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
yuesong: Ian! Didn't expect to see your comment. How are you!

tedpearson: @jhorwit2 - I changed disk metrics to use percent used instead of free bytes. Should make more sense across the heterogeneous cluster now. @tea-dragon - I'm planning some rework of the balancing system, specifically allowing rebalancing when busy, not considering max disk/task usage in hostscore, and more, including things like not waiting until all N rebalances finish before kicking off additional ones.

tea-dragon: @yuesong - better now, thanks.
@tedpearson - I tried to think of some relevant advice, but there is just too much. So... good luck.

tedpearson:
It's spawn town.

jhorwit2:
gltm

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -172,9 +172,9 @@ public SpawnBalancer(Spawn spawn, HostManager hostManager) {
...
     private void initMetrics() {
-        SpawnBalancer.makeGauge("minDiskFreeDiff", () -> minDiskFreeDiff);
-        SpawnBalancer.makeGauge("maxDiskFreeDiff", () -> maxDiskFreeDiff);
-        SpawnBalancer.makeGauge("avgDiskFreeDiff", () -> avgDiskFreeDiff);
+        SpawnBalancer.makeGauge("minDiskPercentUsedDiff", () -> minDiskPercentUsedDiff);
+        SpawnBalancer.makeGauge("maxDiskPercentUsedDiff", () -> maxDiskPercentUsedDiff);
+        SpawnBalancer.makeGauge("avgDiskPercentUsedDiff", () -> avgDiskPercentUsedDiff);
         SpawnBalancer.makeGauge("minTaskPercentDiff", () -> minTaskPercentDiff);
         SpawnBalancer.makeGauge("minTaskPercentDiff", () -> minTaskPercentDiff);
         SpawnBalancer.makeGauge("minTaskPercentDiff", () -> minTaskPercentDiff);
```
yuesong: copy pasta error

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1207,52 +1207,47 @@ public boolean isDiskFull(HostState host) {
...
             for (HostState host : hosts) {
-                long diskDiff = Math.abs(avgDiskFree - host.getAvailDiskBytes());
-                sumDiskFreeDiff += diskDiff;
+                HostScore score = calculateHostScore(host, maxMeanActive, maxDiskPercentUsed);
+                cachedHostScores.put(host.getHostUuid(), score);
+
+                // update average metrics
+                double diskDiff = Math.abs(avgDiskPercentUsed - getUsedDiskPercent(host));
+                sumDiskPercentUsedDiff += diskDiff;
                 double taskDiff = cachedHostScores.get(host.getHostUuid()).getScoreValue(false);
```
yuesong: Just use `score` instead of `cachedHostScores.get(host.getHostUuid())`

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
tedpearson: Oops, fixed

PR #235, opened 2017-04-21
Rebalance when busy, make min free disk absolute

tedpearson:
NOTE: This pull request builds on the already existing pull request on the rebalance branch.

These are some quick fixes to make rebalancing keep up better and clean up some code.

change disk free space to force rebalance or not allow new tasks from percent to absolute numbers
change host score equation to not include cluster max values, each host’s score is independent
move some methods inside of hostscore: getDiskUsedPercent, getAvailDiskBytes
Refs T66734


tedpearson:
Abandoning this to make a new one with less commits and actually compare to current master

PR #236, opened 2017-04-26
changed fs-dead warning message and buttons on host menu ref T64935

satang500:


satang500:
it's included in gui branch

PR #237, opened 2017-04-26
Chooseone

satang500:
Not allow failing more than one minion at a time fs-dead ref T64943


hydra-web/src/js/modules/datatable.js
```diff
@@ -921,6 +936,7 @@ function(_, $, dt, Backbone){
             localStorage[this.id+"_selected"]=JSON.stringify(this.selectedIds);
             var keys= _.keys(this.selectedIds), checkbox=this.views.selectable.find("span.checkbox");
             this.views.parent.find("#selectedCount").html("["+keys.length+"]");
+
```
tedpearson: It looks like you first tried working in this section of code. I think that was right, and I have some advice for a potential better way to implement only enabling #hostFailFsDeadButton if one row is selected. Instead of a change to handleSelect,  I think a change to saveSelectedState is the proper way to handle it.

Specifically:
- Add a class to #hostFailFsDeadButton like "btn-show-one"
- Add handling in saveSelectedState for "btn-show-one" that removes class "disabled" if keys.length=1, and adds class "disabled" if keys.length != 1

I applaud your efforts in learning backbone and javascript. The code you've written probably works great, but I think the approach I'm suggesting will result in cleaner and more maintainable code in the long run.

hydra-web/src/js/modules/host.js
```diff
@@ -224,8 +224,8 @@ function(
 					alertify.error("Fatal warning for failing " + uuids + ": " + data.fatal + "; fail aborted");
 					return;
 				}
-				var msg = "Are you sure you want to fail " + uuids + "?\n";
-				msg += "After failing, cluster will go from " + util.formatPercent(data.prefail) + "% disk used to " + util.formatPercent(data.postfail) + "%.\n";	
+				var msg = "<span class=\"bold-red\">Do you seriously want to fail (file system dead) " + uuids + "?</span><br/>";
+				msg += "(After failing, cluster will go from " + util.formatPercent(data.prefail) + "% disk used to " + util.formatPercent(data.postfail) + "%)";
```
tedpearson: I know you made two pull requests, but I'm going to comment on all issues just in this one since it has both commits.

This change here will result in both failing fs-dead and fs-okay to both say "Do you seriously want to fail (file system dead)" and that's not correct. Please make use of the `deadFs` variable passed into this function to fix this.

hydra-web/src/templates/host.selectable.html
```diff
@@ -19,15 +19,15 @@
...
-    <button id="hostFailFsDeadButton" class='btn btn-default btn-red btn-hide-zero'>
-        Fail (filesystem dead)
+    <button id="hostFailFsOkayButton" class='btn btn-default btn-blue btn-hide-zero'>
+        Queue to Fail
     </button>
     <button id="hostFailCancelButton" class='btn btn-default btn-blue btn-hide-zero'>
         Cancel failure
     </button>
+    <button id="hostFailFsDeadButton" class='btn btn-default btn-red btn-hide-zero'>
+        Fail Now!
```
tedpearson: I really really like the change to the button text. Well done.

hydra-web/src/js/modules/datatable.js
```diff
@@ -949,7 +965,6 @@ function(_, $, dt, Backbone){
             if(!$(event.currentTarget).is("button.checkbox-button")){
                 return;
             }
-            var button = this.views.parent.find("button.checkbox-button");
```
satang500: hum, it was mistakenly removed. nice catch.

hydra-main/web/spawn2/build/1.main.js
```diff
@@ -23542,11 +23542,17 @@ webpackJsonp([1],Array(215).concat([
...
-	                checkbox.parent().data("state","partial");
-	                this.views.selectable.find("button.btn-hide-zero").removeClass("disabled");
+	                if(keys.length == 1) {
+	                    this.views.selectable.find("button.btn-show-one").removeClass("disabled");
+	                } else {
+	                    checkbox.attr("class","checkbox checked-partial");
+	                    checkbox.parent().data("state","partial");
+	                    this.views.selectable.find("button.btn-hide-zero").removeClass("disabled");
+	                    this.views.selectable.find("button.btn-show-one").addClass("disabled");
+	                }
```
yuesong: It's slightly better to flatten this. Instead of:
```
else if (keys.length>0) {
  if (keys.length == 1) {
    ...
  } else {
    ...
  }
}
```

just do:
```
else if (keys.length==1) {
  ...
} else if (keys.length>1) {
  ...
}
```

hydra-web/src/js/modules/datatable.js
```diff
@@ -925,11 +925,17 @@ function(_, $, dt, Backbone){
...
                 this.views.selectable.find("button.btn-hide-zero").removeClass("disabled");
+                this.views.selectable.find("button.btn-show-one").addClass("disabled");
             }
             else if(keys.length>0){
-                checkbox.attr("class","checkbox checked-partial");
-                checkbox.parent().data("state","partial");
-                this.views.selectable.find("button.btn-hide-zero").removeClass("disabled");
+                if(keys.length == 1) {
+                    this.views.selectable.find("button.btn-show-one").removeClass("disabled");
+                } else {
```
tedpearson: Moving this code into an else will cause other btn-hide-zero buttons to not become enabled when there is only 1 selected item in the table.

hydra-web/src/templates/host.selectable.html
```diff
@@ -19,15 +19,15 @@
...
-    </button>
-    <button id="hostFailFsDeadButton" class='btn btn-default btn-red btn-hide-zero'>
-        Fail (filesystem dead)
+    <button id="hostFailFsOkayButton" class='btn btn-default btn-blue btn-hide-zero'>
+        Queue to Fail
     </button>
     <button id="hostFailCancelButton" class='btn btn-default btn-blue btn-hide-zero'>
         Cancel failure
     </button>
+    <button id="hostFailFsDeadButton" class='btn btn-default btn-red btn-hide-zero btn-show-one'>
```
tedpearson: I would recommend against leaving btn-hide-zero on this button, because that implies certain behavior, and it's not obvious what the business logic is. Instead, handle the two types of buttons separately in the javascript.

hydra-web/src/js/modules/datatable.js
```diff
@@ -925,11 +925,17 @@ function(_, $, dt, Backbone){
                 checkbox.attr("class","checkbox checked-all");
                 checkbox.parent().data("state","all");
                 this.views.selectable.find("button.btn-hide-zero").removeClass("disabled");
+                this.views.selectable.find("button.btn-show-one").addClass("disabled");
```
tedpearson: I'm not quite sure when the code in this if statement runs. Do you know what exactly the condition `if(_.isEqual(keys.length,this.collection.length))` is checking here?

hydra-web/src/templates/host.selectable.html
satang500: I didn't separate btn-show-one button and used the current code as much as I can. If I separate it I think I have to add another module to select all button other than "Fail Now!" button.

satang500:
it's included in gui branch

PR #238, opened 2017-04-26
Make Job ID link to job config on hydra alias tab ref T67443

satang500:


hydra-web/src/js/modules/alias.js
```diff
@@ -131,7 +131,7 @@ function(
                     "bVisible":true,
                     "bSearchable":true,
                     "mRender":function(val,type,data){
-                        return val.join(",");
+                        return "<a href='#jobs/"+encodeURIComponent(val)+"/conf'>"+val+"</a>";
```
tedpearson: `val` appears to be an array (at least some times) since it's being joined in the original code. Have you been able to identify if it's an array and if it's ever more than one element?

satang500: If there are multiple job ids you can see one long string which is appending ids, I've searched an alias in spawn prod., but couldn't find one which has multiple ids. But if there is a case where using multiple ids, returning a list of ids which can click seperate ids might be one of possible approaches. BTW, I'm wondering the scenario when it's using muptile id for an alias.

yuesong: Hydra allows multiple job ids to be mapped to the same alias. It was for an old use case. Although it's not used anymore, the right thing to do is to generate a separate job link for each element in the array.

tedpearson: The key point is, when dealing with an array, at the very least you should unwrap it by getting just the first element instead of just using the array directly.

satang500:
it's included in gui branch

PR #239, opened 2017-04-26
Rebalance when busy, make min free disk absolute

tedpearson:
These are some quick fixes to make rebalancing keep up better and clean up some code.

- change disk free space to force rebalance or not allow new tasks from percent to absolute numbers
- change host score equation to not include cluster max values, each host’s score is independent
- move some methods inside of hostscore: getDiskUsedPercent, getAvailDiskBytes

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/job/mq/HostState.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/mq/HostState.java
@@ -25,6 +25,7 @@ import java.util.Set;
 import com.addthis.hydra.minion.Minion;
 
 import com.google.common.base.Objects;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
@@ -403,10 +404,17 @@ public class HostState implements HostMessage {
     }
 
     public double getDiskUsedPercentModified(long reduce) {
+        Preconditions.checkArgument(reduce >= 0, "reduce must be non-negative");
         if ((max == null) || (used == null) || (max.getDisk() <= 0)) {
             return 0;
         }
-        return (double) (used.getDisk() / (max.getDisk() - reduce));
+        double value = (double) (used.getDisk() / (max.getDisk() - reduce));
+        if (value < 0) {
+            return 0;
+        } else if (value > 1) {
+            return 1;
+        }
+        return value;
     }
 
     public void setDead(boolean dead) {
--- a/hydra-main/src/main/java/com/addthis/hydra/job/mq/HostState.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/mq/HostState.java
@@ -391,6 +391,24 @@ public class HostState implements HostMessage {
         this.max = max;
     }
 
+    public long getAvailDiskBytes() {
+        if ((max == null) || (used == null)) {
+            return 1; // Fix some tests
+        }
+        return max.getDisk() - used.getDisk();
+    }
+
+    public double getDiskUsedPercent() {
+        return getDiskUsedPercentModified(0);
+    }
+
+    public double getDiskUsedPercentModified(long reduce) {
+        if ((max == null) || (used == null) || (max.getDisk() <= 0)) {
+            return 0;
+        }
+        return (double) (used.getDisk() / (max.getDisk() - reduce));
+    }
+
     public void setDead(boolean dead) {
         this.dead = dead;
     }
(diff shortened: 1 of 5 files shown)


hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancerConfig.java
```diff
@@ -72,10 +72,10 @@
     // Track the last time a job autobalance was done
     private long lastJobAutobalanceTime = 0L;
 
-    // If you're at more than 80% disk capacity, you don't get assigned any new tasks or replicas
-    private double minDiskPercentAvailToReceiveNewTasks = Double.parseDouble(Parameter.value("spawnbalance.min.disk.percent.avail.newtasks", ".2"));
-    // If you're at more than 90% disk capacity, push tasks off before running additional tasks
-    private double minDiskPercentAvailToRunJobs = Double.parseDouble(Parameter.value("spawnbalance.min.disk.percent.avail.runtasks", ".1"));
+    // If you have less than 700GB free disk space, push tasks off before running additional tasks
```
yuesong: Comment should not be the same as that of the parameter below?

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1256,26 +1243,18 @@ protected void updateAggregateStatistics(List<HostState> hosts) {
...
-        score += activeTaskWeight * Math.pow(meanActive / clusterMaxMeanActive, 2.5);
-        score += diskUsedWeight * Math.pow(usedDiskPercent / clusterMaxDiskUsed, 2.5);
+        double exponent = 2.5;
+        double score = activeTaskWeight * Math.pow(meanActive, exponent);
+        double diskPercentPowered = Math.pow(diskUsedPercentModified, exponent);
+        score += diskUsedWeight * diskPercentPowered;
         // If host is very full, make sure to give the host a big score
-        score = Math.max(score, (activeTaskWeight + diskUsedWeight) * usedDiskPercent);
-        return new HostScore(meanActive, usedDiskPercent, score);
+        score = Math.max(score, (activeTaskWeight + diskUsedWeight) * diskPercentPowered);
```
yuesong: Old code uses `usedDiskPercent` while new code uses `diskPercentPowered`. Is that intentional?

hydra-main/src/main/java/com/addthis/hydra/job/mq/HostState.java
```diff
@@ -391,6 +391,24 @@ public void setMax(HostCapacity max) {
...
+
+    public double getDiskUsedPercent() {
+        return getDiskUsedPercentModified(0);
+    }
+
+    public double getDiskUsedPercentModified(long reduce) {
+        if ((max == null) || (used == null) || (max.getDisk() <= 0)) {
+            return 0;
+        }
+        return (double) (used.getDisk() / (max.getDisk() - reduce));
```
yuesong: Should check return value to make sure it's in the range of [0, 1]. Also add javadoc here and/or comment where it is called to explain usage/reasoning

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancerConfig.java
tedpearson: I was hoping I could sneak this through and fix it later 😛

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
tedpearson: Old code was wrong?

tedpearson:
Comments addressed

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
yuesong: I still don't quite understand how it gives host with fuller disk a higher score.

The default configuration is: activeTaskWeight=30 and diskUsedWeight=70. meanActive is about 2 and doesn't seem to change a lot over time, according to ganglia metric. Now let's say disk is full, so diskUsedPercentModified=1

The original formula (before the Math.max call) produces: score=30*pow(2, 2.5) + 70*pow(1, 2.5)=170+70=240
The alternative formula (second parameter in the Math.max call) produces: score=(30+70)*pow(1, 2.5)=100

The original score still wins.

The thing is: disk percent is <=1 while mean active seems to be around 2. With the exponential, mean active is almost always going to be the dominant factor.

Where am I wrong in this?

tedpearson:
You are wrong in that `host.getMeanActiveTasks()` returns a value from 0 to 1, and would never return 2. See Minion.java line 660.

yuesong:
Good enough.

PR #243, opened 2017-05-16
reduce default disk free; update unit tests

tedpearson:
Reducing default disk space required to 10/20 GB to run tasks or receive new ones. Configure your local environment differently if desired.


yuesong:
Is this to make local_stack works better with the default values? Will hydra-addthis config have the larger values?

yuesong:
never mind. I saw your diff.

PR #246, opened 2017-05-23
removed huge error when kick the disabled job ref T64087

satang500:


yuesong:
Two issues with this fix: 1) it assumes that job kick may fail only because the job is disabled. That is not true. 2) This does not address the case where a user kicks a disabled job from the job details screen.

The right fix requires more than a simple change of the error message in the front end. Please review sever side code in `JobsResource.startJob`, then we can pair to figure out the right approach.

yuesong:
@satang500 please close this PR

PR #248, opened 2017-05-26
prevent a job creation without creator ref T68980

satang500:


yuesong:
Please close this PR as another is open to address the same issue using a different approach

yuesong:
@satang500 Please close this PR

PR #250, opened 2017-06-05
Kickdis1

satang500:


hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -1363,24 +1370,17 @@ public Response saveAllJobs(@QueryParam("user") String user,  @QueryParam("token
             log.trace("Save all jobs exception", ex);
             return Response.ok("{\"operation\":\"failed: " + ex.toString() + "\"").build();
         }
-
     }
 
     private static Response buildServerError(Exception exception) {
-        log.warn("", exception);
+        log.error("Error in kick job: {}", Throwables.getStackTraceAsString(exception));
```
yuesong: Do not use `Throwables.getStackTraceAsString` to parameterize a logging message. The correct methods on slf4j logger for logging an exception are:
```
public void warn(String msg, Throwable t);
public void error(String msg, Throwable t);
```

hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -1363,24 +1370,17 @@ public Response saveAllJobs(@QueryParam("user") String user,  @QueryParam("token
...
-
-        final String response = "{" +
-                "\"error\": \"A java exception was thrown." +
-                "\"message\": \"" + StringEscapeUtils.escapeEcmaScript(message) + "\"" +
-                "\"stack\": \"" + StringEscapeUtils.escapeEcmaScript(stack) + "\"" +
-                "\"}";
-
+        final String response =
+                "ERROR: " + exception.getClass() + " was thrown <br/>" +
+                "MESSAGE: " + StringEscapeUtils.escapeEcmaScript(message);
```
yuesong: `buildServerError` is used in many places. Have you done a proper audit of all the js code where the returned server error is used? If not, the change from json to plain text can potentially break front end things. A safer change is to keep the json format, but remove the `stack` element - the stack trace would almost certainly mess up fron end display and should not be shown to the end user anyway - so this should be an improvement

hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -1016,6 +1019,8 @@ public Response startJob(@QueryParam("jobid") Optional<String> jobIds,
                 IJob job = spawn.getJob(jobId);
```
yuesong: This block of code of validating and kicking a job is identical to the code above. They should be refactored into a method to be used in both places

yuesong:
The new code makes JobsResource not thread-safe. Let's pair on this.

hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -1372,14 +1373,10 @@ private static Response buildServerError(Exception exception) {
...
-        String stack = Throwables.getStackTraceAsString(exception);
-
         final String response = "{" +
-                "\"error\": \"A java exception was thrown." +
-                "\"message\": \"" + StringEscapeUtils.escapeEcmaScript(message) + "\"" +
-                "\"stack\": \"" + StringEscapeUtils.escapeEcmaScript(stack) + "\"" +
-                "\"}";
+                                "\"error\": \"A java exception was thrown." +
+                                "\"message\": \"" + StringEscapeUtils.escapeEcmaScript(message) + "\"" +
+                                "\"}";
```
yuesong: Hmm... Is the intent to respond with json text? This is not valid json (closing quote for the error message text is missing, so is the comma). This looks to be an existing bug, but we should fix regardless.

hydra-web/src/js/modules/jobs.js
```diff
@@ -77,10 +77,13 @@ function(
...
         if (data.success.length > 0) {
-            alertify.success(data.success.length + " job(s) have been " + state, 5);
+            alertify.success(data.success.length + " job(s) have been " + state, 10);
         }
         if (data.error.length > 0) {
-            alertify.error(data.error.length + " job(s) have not been " + state, 5);
+            alertify.error(data.error.length + " job(s) have not been " + state, 10);
+        }
+        if (data.disable && data.disable.length > 0) {
+            alertify.error(data.disable.length + " job(s) have been disabled " + state, 10);
```
yuesong: The error message should just be:
```
data.disable.length + " job(s) are disabled "
```
`+ state` doesn't make sense in this case because its value is `"started"`

hydra-web/src/js/modules/jobs.js
```diff
@@ -77,10 +77,13 @@ function(
...
         if (data.success.length > 0) {
-            alertify.success(data.success.length + " job(s) have been " + state, 5);
+            alertify.success(data.success.length + " job(s) have been " + state, 10);
         }
         if (data.error.length > 0) {
-            alertify.error(data.error.length + " job(s) have not been " + state, 5);
+            alertify.error(data.error.length + " job(s) have not been " + state, 10);
+        }
+        if (data.disable.length > 0) {
+            alertify.error(data.disable.length + " job(s) are disabled ", 10);
```
yuesong: Delete trailing space in message

PR #252, opened 2017-06-08
moved code to different places

satang500:


hydra-main/src/test/java/com/addthis/hydra/job/web/JobRequestHandlerImplTest.java
```diff
@@ -123,6 +121,17 @@ public void createJob_ConfigTooLarge() throws Exception {
     }
 
     @Test
+    public void createJob_WithoutCreator() throws Exception {
+        Job job = new Job();
+        when(spawn.createJob("megatron", -1, Collections.<String> emptyList(), "default", "default-task", false)).thenReturn(job);
+
+        kv.add("config", "my job config");
+        kv.add("command", "default-task");
```
yuesong: You are testing the case where the front end passes empty string as creator. You should set up `kv` properly to simulate that:
```
kv.add("creator", "")
```
Accordingly, the stub for `Spawn.createJob` should be changed too because you are expecting an empty first argument which is creator:
```
when(spawn.createJob("", -1, Collections.<String> emptyList(), "default", "default-task", false)).thenReturn(job);
```

hydra-main/src/main/java/com/addthis/hydra/job/web/JobRequestHandlerImpl.java
```diff
@@ -78,7 +78,7 @@ public Job createOrUpdateJob(KVPairs kv, String username, String token, String s
                     Splitter.on(',').omitEmptyStrings().trimResults().splitToList(kv.getValue("hosts", "")),
                     kv.getValue("minionType", DEFAULT_MINION_TYPE),
                     command, defaults);
-            updateOwnership(job, user);
+            updateOwnership(kv, job, user);
```
yuesong: Actually I think the best place to do this is before line 75 `job = spawn.createJob(` by setting the "creator" and "owner", and "group" entries in the `kv` if they are null or empty. Then you won't need the `updateOwnership` method anymore.

satang500: so do you also make "owner", and "group" not empty as well?

PR #257, opened 2017-07-10
revert job/task that errored on the first run ref T68696

satang500:


yuesong:
I haven't looked at all the details, but it's clear to me that a better place to make the change is `com.addthis.hydra.minion.JobTask` which contains implementation of the current revert behavior, such as copying backup to live etc. Let's find some time to review the code together.

PR #259, opened 2017-07-13
revert job that errored on the first run (not revert task yet) ref T6…

satang500:
revert task (which requires the modification of gui) is not done yet
if revert job is ok, i will work on revert task as well


tedpearson:
This has been fixed by a different method.

PR #260, opened 2017-07-18
Add an endpoint to change minion type after job created ref T64107

kxingit:
Add an endpoint to change minion type after job created ref T64107

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/job/web/JobRequestHandlerImpl.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/web/JobRequestHandlerImpl.java
@@ -156,12 +156,7 @@ public class JobRequestHandlerImpl implements JobRequestHandler {
     }
 
     private boolean isJobOnMinionType(Job job, String minionType) {
-        // make a valid host set
-        Set<String> hostsWithTargetMinionType = new HashSet<String>();
-        List<HostState> hostStates = spawn.hostManager.listHostStatus(minionType);
-        for (HostState hostState : hostStates) {
-            hostsWithTargetMinionType.add(hostState.getHostUuid());
-        }
+        Set<String> hostsWithTargetMinionType = getHostsWithTargetMninionType(minionType);
 
         // check tasks
         for (JobTask jobTask : job.getCopyOfTasks()) {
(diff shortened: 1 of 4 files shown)


hydra-main/src/main/java/com/addthis/hydra/job/web/JobRequestHandlerImpl.java
```diff
@@ -116,6 +115,70 @@ public Job createOrUpdateJob(KVPairs kv, String username, String token, String s
...
+        // make a valid host set
+        Set<String> hostsWithTargetMinionType = new HashSet<String>();
+        List<HostState> hostStates = spawn.hostManager.listHostStatus(minionType);
+        for (HostState hostState : hostStates) {
+            hostsWithTargetMinionType.add(hostState.getHostUuid());
+        }
+
+        // check tasks and their replicas
+        for (JobTask jobTask : job.getCopyOfTasks()) {
+            if (!hostsWithTargetMinionType.contains(jobTask.getHostUUID())) {
```
yuesong: Is `jobTask.getHostUUID` not one of `jobTask.getReplicas`?

hydra-main/src/main/java/com/addthis/hydra/job/web/resources/JobsResource.java
```diff
@@ -779,7 +779,7 @@ public Response saveJob(@QueryParam("pairs") KVPairs kv,
 
     /**
      *
-     * @param jobId         job id
+     * @param job           job id
```
kxingit: right, I was looking at `@QueryParam("job")`, that was wrong and will be fixed in the next push.

hydra-main/src/main/java/com/addthis/hydra/job/web/JobRequestHandlerImpl.java
kxingit: I read the code, but did not fully understand weather it is the case. So I ran a remote debugger for a job with `number of replicas` set to 2, in which case there are 1+2 copies in total. The debugger shows there are two replicas. I think this means `jobTask.getHostUUID` is not in `jobTask.getReplicas`. I will further read the code to confirm.

kxingit:
Could not figure out a way to make streams/lambda more readable than a for loop. So left unchanged for now. Unit tests added, in the next commit.

hydra-main/src/test/java/com/addthis/hydra/job/web/JobRequestHandlerImplTest.java
```diff
@@ -117,33 +117,22 @@ public void updateMinionType_notOnTargetType() throws Exception {
...
             // Expected
         }
-        verify(spawn, times(0)).updateJob(job);
-        assertSame(job.getMinionType(), "oldMinion");
+        verify(spawn, never()).updateJob(job);
+        assertEquals("oldMinion", job.getMinionType());
     }
 
-    @Test
+    @Test(expected = IllegalArgumentException.class)
```
yuesong: This is the convention for verifying an expected exception. I also simplified the test. I think it's a better balance between test coverage and simplicity.

PR #263, opened 2017-07-26
add murmur hash to ValueFilterHash ref T69516

satang500:


yuesong:
- use Murmur3HashFunction
- update the javadoc

satang500:
murmur3_128() is implementing 128-bit murmur3 algorithm, x64 variant. Are you referring to something else?

satang500:
murmur3 is already implemented ValueFilterHash. use type 1

yuesong:
Good observation. The kicker is this: type 1 uses `PluggableHashFunction` and the underlying hashing function is configurable (or pluggable, hence the name). In hydra-oss, the configuration is in `hydra/hydra-essentials/src/main/resources/reference.conf`, where the default is `Murmur3HashFunction`. This is overridden in hydra-addthis by `hydra-addthis/hydra-essentials-internal/src/main/resources/application.conf` - `SuperFastHashFunction` is the default hasher. This is for compatibility reasons as murmur hash was a late addition. The ask here is basically to allow us to explicitly use murmur hash for the filter.

hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/ValueFilterHash.java
```diff
@@ -99,6 +101,10 @@ public ValueObject filterValue(ValueObject value) {
                 } catch (NoSuchAlgorithmException e) {
                     // ignore
                 }
+            case 5:
+                StringAndByteHashFunction mmh = new Murmur3HashFunction();
```
yuesong: It is wasteful to create a new instance of `Murmur3HashFunction` every time `filterValue` is called. A better approach is to use a singleton instance (e.g. it can be a static field of `Murmur3HashFunction` class). The singleton can also be used in `PluggableHashFunction`.

satang500: StringAndByteHashFunction is a interface and implemented by some classes. To make methods in Murmur3HashFunction static, we need also make methods in its parent also static, but interface can't have static method in java7. In java8, we can have static methods in an interface, but  we can’t override them in the implementation classes.  If you want to make all method calls in ValueFilterHash static, need to change structure of related classes using pattern (eg factory pattern).

satang500:
merged to master

PR #265, opened 2017-08-10
fix light host rebalance and add unit test

apsubra:
moveLimit parameter added to limit tasks moved to light hosts after rebalance. Also fixed unit test (overloaded host rebalance) to avoid moving tasks to readonly hosts.

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
@@ -14,9 +14,9 @@
 package com.addthis.hydra.job.spawn.balancer;
 
 import java.util.ArrayList;
-import java.util.List;
 
 import com.addthis.hydra.job.JobTaskMoveAssignment;
+import com.addthis.hydra.job.mq.JobKey;
 import com.addthis.hydra.job.spawn.Spawn;
 
 class MoveAssignmentList extends ArrayList<JobTaskMoveAssignment> {
@@ -38,10 +38,13 @@ class MoveAssignmentList extends ArrayList<JobTaskMoveAssignment> {
         return super.add(assignment);
     }
 
-    public void addAll(List<JobTaskMoveAssignment> assignments){
-        for(JobTaskMoveAssignment assignment: assignments) {
-            add(assignment);
+    public boolean contains(JobKey jobKey) {
+        for (JobTaskMoveAssignment assignment : this) {
+            if(assignment.getJobKey().getJobUuid().equals(jobKey.getJobUuid())) {
+                return true;
+            }
         }
+        return false;
     }
 
     public long getBytesUsed() {
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
@@ -21,7 +21,6 @@ import com.addthis.hydra.job.spawn.Spawn;
 
 class MoveAssignmentList {
     private List<JobTaskMoveAssignment> moveAssignmentList;
-    private static final long serialVersionUID = -563719566151798849L;
 
     private final Spawn spawn;
     private final SpawnBalancerTaskSizer taskSizer;
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
@@ -14,6 +14,7 @@
 package com.addthis.hydra.job.spawn.balancer;
 
 import java.util.ArrayList;
+import java.util.List;
 
 import com.addthis.hydra.job.JobTaskMoveAssignment;
 import com.addthis.hydra.job.spawn.Spawn;
@@ -35,6 +36,12 @@ class MoveAssignmentList extends ArrayList<JobTaskMoveAssignment> {
     public boolean add(JobTaskMoveAssignment assignment) {
         bytesUsed += taskSizer.estimateTrueSize(spawn.getTask(assignment.getJobKey()));
         return super.add(assignment);
+    }
+
+    public void addAll(List<JobTaskMoveAssignment> assignments){
+        for(JobTaskMoveAssignment assignment: assignments) {
+            add(assignment);
+        }
     }
 
     public long getBytesUsed() {
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
@@ -15,8 +15,10 @@ package com.addthis.hydra.job.spawn.balancer;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import com.addthis.hydra.job.JobTaskMoveAssignment;
+import com.addthis.hydra.job.mq.JobKey;
 import com.addthis.hydra.job.spawn.Spawn;
 
 class MoveAssignmentList {
@@ -42,26 +44,27 @@ class MoveAssignmentList {
         return this.moveAssignmentList.isEmpty();
     }
 
-    /**
-     * Returns <tt>true</tt> if the list contains a move assignment
-     * to move any task associated with the specified job.
-     * @param jobId JobID
-     * @return <tt>true</tt> if the list contains an assignment moving any of the job's tasks
-     */
-    public boolean containsJob(String jobId) {
-        if (jobId == null) {
-            return false;
-        }
-        return moveAssignmentList.stream().anyMatch(assignment -> jobId.equals(assignment.getJobKey().getJobUuid()));
-    }
-
     public List<JobTaskMoveAssignment> getList() {
         return this.moveAssignmentList;
     }
 
+    /**
+     * Add the given JobTaskMoveAssignment to the moveAssignmentList
+     * if the assignment does not move a replica of the same task to the same target host
+     * @param assignment JobTaskMoveAssignment to be added to the moveAssignmentList
+     * @return <tt>true</tt> if this JobTaskMoveAssignment was added to the moveAssignmentList
+     */
     public boolean add(JobTaskMoveAssignment assignment) {
-        bytesUsed += taskSizer.estimateTrueSize(spawn.getTask(assignment.getJobKey()));
-        return this.moveAssignmentList.add(assignment);
+        List<JobKey> jobKeysToTargetHost = moveAssignmentList.stream()
+                          .filter(moveAssignment -> moveAssignment.getTargetUUID().equals(assignment.getTargetUUID()))
+                          .map(moveAssignment -> moveAssignment.getJobKey())
+                          .collect(Collectors.toList());
+
+        if(!jobKeysToTargetHost.contains(assignment.getJobKey())) {
+            bytesUsed += taskSizer.estimateTrueSize(spawn.getTask(assignment.getJobKey()));
+            return this.moveAssignmentList.add(assignment);
+        }
+        return false;
     }
 
     public long getBytesUsed() {
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
@@ -15,10 +15,8 @@ package com.addthis.hydra.job.spawn.balancer;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
 
 import com.addthis.hydra.job.JobTaskMoveAssignment;
-import com.addthis.hydra.job.mq.JobKey;
 import com.addthis.hydra.job.spawn.Spawn;
 
 class MoveAssignmentList {
@@ -55,14 +53,16 @@ class MoveAssignmentList {
      * @return <tt>true</tt> if this JobTaskMoveAssignment was added to the moveAssignmentList
      */
     public boolean add(JobTaskMoveAssignment assignment) {
-        List<JobKey> jobKeysToTargetHost = moveAssignmentList.stream()
-                          .filter(moveAssignment -> moveAssignment.getTargetUUID().equals(assignment.getTargetUUID()))
-                          .map(moveAssignment -> moveAssignment.getJobKey())
-                          .collect(Collectors.toList());
+        // Check if moveAssignmentList contains an assignment that moves a replica of the task to the same target host
+        boolean isJobKeyPresent = this.moveAssignmentList.stream()
+                                                         .anyMatch(moveAssignment ->
+                                                                           moveAssignment.getJobKey().equals(assignment.getJobKey()) &&
+                                                                           moveAssignment.getTargetUUID().equals(assignment.getTargetUUID()));
 
-        if(!jobKeysToTargetHost.contains(assignment.getJobKey())) {
+        if(!isJobKeyPresent) {
             bytesUsed += taskSizer.estimateTrueSize(spawn.getTask(assignment.getJobKey()));
-            return this.moveAssignmentList.add(assignment);
+            this.moveAssignmentList.add(assignment);
+            return true;
         }
         return false;
     }
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
@@ -45,16 +45,14 @@ class MoveAssignmentList {
     /**
      * Returns <tt>true</tt> if the list contains a move assignment
      * to move any task associated with the specified job.
-     * @param jobID JobID
+     * @param jobId JobID
      * @return <tt>true</tt> if the list contains an assignment moving any of the job's tasks
      */
-    public boolean containsJob(String jobID) {
-        for (JobTaskMoveAssignment assignment : this.moveAssignmentList) {
-            if(assignment.getJobKey().getJobUuid().equals(jobID)) {
-                return true;
-            }
+    public boolean containsJob(String jobId) {
+        if (jobId == null) {
+            return false;
         }
-        return false;
+        return moveAssignmentList.stream().anyMatch(assignment -> jobId.equals(assignment.getJobKey().getJobUuid()));
     }
 
     public List<JobTaskMoveAssignment> getList() {
--- a/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
@@ -14,12 +14,13 @@
 package com.addthis.hydra.job.spawn.balancer;
 
 import java.util.ArrayList;
+import java.util.List;
 
 import com.addthis.hydra.job.JobTaskMoveAssignment;
-import com.addthis.hydra.job.mq.JobKey;
 import com.addthis.hydra.job.spawn.Spawn;
 
-class MoveAssignmentList extends ArrayList<JobTaskMoveAssignment> {
+class MoveAssignmentList {
+    private List<JobTaskMoveAssignment> moveAssignmentList;
     private static final long serialVersionUID = -563719566151798849L;
 
     private final Spawn spawn;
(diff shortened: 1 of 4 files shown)


yuesong:
Haven't reviewed everything yet, but `MoveAssignmentList.addAll` doesn't update bytesUsed as you might have assumed, because the base implementation of that method in `ArrayList` doesn't use `add`.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1403,15 +1403,38 @@ private JobTaskItemByHostMap generateTaskCountByHost(List<HostState> hosts, Iter
...
+    private List<JobTaskMoveAssignment> pushTasksOntoHost(HostState host, Iterable<HostState> heavyHosts,
+                                                          boolean limitBytes, double byteLimitFactor, int moveLimit,
+                                                          boolean obeyDontAutobalanceMe) {
+
         MoveAssignmentList moveAssignments = new MoveAssignmentList(spawn, taskSizer);
+        List<JobTaskMoveAssignment> rv = purgeMisplacedTasks(host, moveLimit);
+        moveAssignments.addAll(rv);
+
+        // Available bytes after purgeMisplacedTasks movements
+        long byteLimit = (long) ((byteLimitFactor * config.getBytesMovedFullRebalance()) - moveAssignments.getBytesUsed());
```
jhorwit2: This is duplicate code from above. Can the useful bits be moved to a shared method?

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -420,30 +420,32 @@ public boolean canReceiveNewTasks(HostState host) {
...
-                if (rv.size() >= moveLimit) {
-                    break;
-                }
+        MoveAssignmentList moveAssignments = new MoveAssignmentList(spawn, taskSizer);
+        moveAssignments.addAll(rv);
+
+        long byteLimit = (long) ((byteLimitFactor * config.getBytesMovedFullRebalance()) - moveAssignments.getBytesUsed());
+        List<HostState> hostsSorted = sortHostsByDiskSpace(otherHosts);
+
+        for (JobTask task : findTasksToMove(host, obeyDontAutobalanceMe)) {
```
jhorwit2: Should you sort the tasks to move? Right now the order seems fairly random. It seems to me that we should try to move as few tasks as possible to avoid the `moveLimit`, which means we would probably want to sort the tasks in descending order by size before iterating over them.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1039,8 +1040,7 @@ private void executeHostFailureRecovery(JobTask task, String failedHostUuid, Cor
             log.info("[spawn.balancer] {} categorized as underloaded host; looking for tasks to pull onto it", hostID);
             List<HostState> heavyHosts =
                     Lists.reverse(sortedHosts.subList(sortedHosts.size() - numAlleviateHosts, sortedHosts.size()));
-            // FIXME: pushTasksOntoHost needs to take a moveLimit parameter
-            // rv.addAll(pushTasksOntoHost(host, heavyHosts));
+            rv.addAll(pushTasksOntoHost(host, heavyHosts, true, 1, config.getTasksMovedFullRebalance(), true));
```
jhorwit2: nit: comment for magic number 1

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1403,15 +1403,38 @@ private JobTaskItemByHostMap generateTaskCountByHost(List<HostState> hosts, Iter
         return rv;
     }
 
-    private List<JobTaskMoveAssignment> pushTasksOntoHost(HostState host, Iterable<HostState> heavyHosts) {
+    private List<JobTaskMoveAssignment> pushTasksOntoHost(HostState host, Iterable<HostState> heavyHosts,
```
jhorwit2: `moveLimit` seems to be taken from the `config` object. You can remove that argument and reference `config` directly since this method is `private` scoped.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
jhorwit2: This is probably out of scope of this PR, but something to consider potentially.

apsubra: Yes, perhaps a better way of choosing tasks to move should be used (Kefeng mentioned this as well). I was concerned that moving only the heaviest task from each heavyHost onto the lightHost could make the light host 'heavy' for the next rebalance. This would be less likely to happen with a random selection of tasks. Thoughts on this?

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1404,15 +1404,26 @@ private JobTaskItemByHostMap generateTaskCountByHost(List<HostState> hosts, Iter
...
+
         for (HostState heavyHost : heavyHosts) {
-            double byteLimitFactor =
-                    1 - ((double) moveAssignments.getBytesUsed() / config.getBytesMovedFullRebalance());
-            moveAssignments.addAll(pushTasksOffHost(heavyHost, Collections.singletonList(host), true, byteLimitFactor,
-                                                    config.getTasksMovedFullRebalance(), true));
+            if (moveAssignments.size() >= moveLimit || !canReceiveNewTasks(host) || moveAssignments.getBytesUsed() >= config.getBytesMovedFullRebalance()) {
+                break;
+            }
+            moveAssignments.addAll(pushTasksOffHost(heavyHost, lightHostList, true, byteLimitFactor, 1, true));
```
yuesong: Did you change the code back to moving multiple tasks from the same heavy host because of my comment on reusing code? I didn't mean reusing `pushTasksOffHost`, just the block of code that are duplicated between `pushTasksOffHost` and `pustTasksOntoHost` in your previous commit. I don't have a strong preference really. I think the idea of pulling tasks from different heave hosts has merit, but it's slightly more complex. I'm not convinced the benefit, if any, is worth that complexity.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
```diff
@@ -37,6 +38,15 @@ public boolean add(JobTaskMoveAssignment assignment) {
         return super.add(assignment);
     }
 
+    public boolean contains(JobKey jobKey) {
```
yuesong: Change to `public boolean containsJob(String jobId)`

I also suggest `MoveAssignmentList` should not extend `ArrayList`. Instead it should have an internal `List` that can be accessed via a `getList` method, and expose the few List methods, such as `add`, `size` etc. that other code needs. In its current form, developers can be misled to make the wrong assumption about its behavior of methods such as `addAll`

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -420,29 +420,31 @@ public boolean canReceiveNewTasks(HostState host) {
...
-                // for the assignment is null
-                if (assignment != null && assignment.getTargetUUID() != null) {
-                    markRecentlyReplicatedTo(assignment.getTargetUUID());
-                    rv.add(assignment);
-                    byteLimit -= taskTrueSize;
-                }
-                if (rv.size() >= moveLimit) {
-                    break;
-                }
+        MoveAssignmentList moveAssignments = new MoveAssignmentList(spawn, taskSizer);
```
yuesong: `MoveAssignmentList.getBytesUsed` isn't used in this method, so I think we can just use a regular `List` to avoid the extra cost of the `add` method. As a matter of fact, the original code just needs minor changes:
- delete first call to `purgeMisplacedTasks`. Create of a new `List<JobTaskMveAssignment>` instead
- get rid of `if (rv.size() <= moveLimit) {` as the new list is empty
- the code within the `if` body remains unchanged
- call purgeMisplacedTasks and add its result to the list before return

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1026,21 +1028,19 @@ private void executeHostFailureRecovery(JobTask task, String failedHostUuid, Cor
...
         int numAlleviateHosts = (int) Math.ceil(sortedHosts.size() * config.getAlleviateHostPercentage());
-
         HostFailWorker.FailState failState = spawn.getHostFailWorker().getFailureState(hostID);
         if ((failState == HostFailWorker.FailState.FAILING_FS_OKAY) || isExtremeHost(hostID, true, true) ||
             (host.getAvailDiskBytes() < config.getMinFreeDiskSpaceToRecieveNewTasks())) {
             // Host disk is overloaded
             log.info("[spawn.balancer] {} categorized as overloaded host; looking for tasks to push off of it", hostID);
             List<HostState> lightHosts = sortedHosts.subList(0, numAlleviateHosts);
-            rv.addAll(pushTasksOffHost(host, lightHosts, true, 1, config.getTasksMovedFullRebalance(), true));
+            rv.addAll(pushTasksOffHost(host, lightHosts, true, 1, config.getTasksMovedFullRebalance(),true));
```
yuesong: Restore space after the last `,`

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1404,15 +1404,33 @@ private JobTaskItemByHostMap generateTaskCountByHost(List<HostState> hosts, Iter
...
+
         for (HostState heavyHost : heavyHosts) {
-            double byteLimitFactor =
-                    1 - ((double) moveAssignments.getBytesUsed() / config.getBytesMovedFullRebalance());
-            moveAssignments.addAll(pushTasksOffHost(heavyHost, Collections.singletonList(host), true, byteLimitFactor,
-                                                    config.getTasksMovedFullRebalance(), true));
+            if (moveAssignments.size() >= moveLimit || !canReceiveNewTasks(host) || moveAssignments.getBytesUsed() >= config.getBytesMovedFullRebalance()) {
+                break;
+            }
+            List<JobTaskMoveAssignment> assignments = pushTasksOffHost(heavyHost, lightHostList,true, byteLimitFactor, 1, true);
```
yuesong: `pushTasksOffHost` calls `markRecentlyReplicatedTo` to mark assigned hosts as recently-replicated-to. But because of the same-job check below, the assignment may be ignored, resulting in incorrect marking of the host.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1404,15 +1403,38 @@ private JobTaskItemByHostMap generateTaskCountByHost(List<HostState> hosts, Iter
...
+            for(JobTaskMoveAssignment assignment : assignments) {
+                if(!moveAssignments.containsJob(assignment.getJobKey().getJobUuid())) {
+                    moveAssignments.add(assignment);
+                    // Update byteLimitFactor with available byte limit
+                    byteLimitFactor = 1 - (moveAssignments.getBytesUsed() / config.getBytesMovedFullRebalance());
+                    targetHosts.add(assignment.getTargetUUID());
+                } else if(!targetHosts.contains(assignment.getTargetUUID())) {
+                    // Target host wrongly marked as recently replicated to
+                    // as we do not accept this moveAssignment
+                    recentlyReplicatedToHosts.invalidate(assignment.getTargetUUID());
```
yuesong: Now that I'm thinking about it, `assignment.getTargetUUID` is always `host` because we are moving tasks to it. Invalidation here is only necessary if we end up not moving any task onto this host. Even in that case, because `recentlyReplicatedToHosts` is a cache with a 15-minute expiration, the host may be in it from a previous rebalance call, so just invalidating it may be incorrect. In other words, you should invalidate only if it is not in `recentlyReplicatedToHosts` previously. At this point, I feel the code is becoming more complex than it needs to be, plus after a quick look, `recentlyReplicatedToHosts` seems to be only marked in `pushTasksOffHost`, which begs the question - why aren't we marking hosts in other places where we move tasks?

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
```diff
@@ -29,12 +31,39 @@
...
+        return this.moveAssignmentList.isEmpty();
+    }
+
+    /**
+     * Returns <tt>true</tt> if the list contains a move assignment
+     * to move any task associated with the specified job.
+     * @param jobID JobID
+     * @return <tt>true</tt> if the list contains an assignment moving any of the job's tasks
+     */
+    public boolean containsJob(String jobID) {
```
yuesong: - Rename `JobID` to `JobId`.
- Check `null`
- And use lambda:
```
        if (jobId == null) {
            return false;
        } else {
            return moveAssignmentList.stream().anyMatch(a -> jobId.equals(a.getJobKey().getJobUuid()));
        }
```

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -411,6 +413,32 @@ public boolean canReceiveNewTasks(HostState host) {
         return pushTasksOffHost(host, hosts, false, 1, moveLimit, false);
     }
 
+    private List<JobTaskMoveAssignment> getMoveAssignments(HostState host, Collection<HostState> otherHosts,
```
yuesong: Why doesn't this method take `moveLimit`? It could exit earlier if it did

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1026,21 +1037,23 @@ private void executeHostFailureRecovery(JobTask task, String failedHostUuid, Cor
...
-
         HostFailWorker.FailState failState = spawn.getHostFailWorker().getFailureState(hostID);
         if ((failState == HostFailWorker.FailState.FAILING_FS_OKAY) || isExtremeHost(hostID, true, true) ||
             (host.getAvailDiskBytes() < config.getMinFreeDiskSpaceToRecieveNewTasks())) {
             // Host disk is overloaded
             log.info("[spawn.balancer] {} categorized as overloaded host; looking for tasks to push off of it", hostID);
             List<HostState> lightHosts = sortedHosts.subList(0, numAlleviateHosts);
-            rv.addAll(pushTasksOffHost(host, lightHosts, true, 1, config.getTasksMovedFullRebalance(), true));
+            List<JobTaskMoveAssignment> moveAssignments = pushTasksOffHost(host, lightHosts, true, 1, config.getTasksMovedFullRebalance(), true);
+            markRecentlyReplicatedTo(moveAssignments);
```
yuesong: I think this is redundant as it is called in `pushTasksOffHost`.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1026,21 +1037,23 @@ private void executeHostFailureRecovery(JobTask task, String failedHostUuid, Cor
...
+            rv.addAll(moveAssignments);
         } else if (isExtremeHost(hostID, true, false)) {
             // Host disk is underloaded
             log.info("[spawn.balancer] {} categorized as underloaded host; looking for tasks to pull onto it", hostID);
             List<HostState> heavyHosts =
                     Lists.reverse(sortedHosts.subList(sortedHosts.size() - numAlleviateHosts, sortedHosts.size()));
-            // FIXME: pushTasksOntoHost needs to take a moveLimit parameter
-            // rv.addAll(pushTasksOntoHost(host, heavyHosts));
+            List<JobTaskMoveAssignment> moveAssignments = pushTasksOntoHost(host, heavyHosts);
+            markRecentlyReplicatedTo(moveAssignments);
```
yuesong: This call is redundant too.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
apsubra: Since pushTasksOntoHost filters out some moveAssignments, this method returns a longer list than necessary. Would you recommend adding a moveLimit and moving fewer tasks onto light hosts (if they're filtered out)?

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1403,16 +1416,35 @@ private JobTaskItemByHostMap generateTaskCountByHost(List<HostState> hosts, Iter
...
+        List<HostState> lightHostList = Arrays.asList(host);
+        double byteLimitFactor = 1;
+
         for (HostState heavyHost : heavyHosts) {
-            double byteLimitFactor =
-                    1 - ((double) moveAssignments.getBytesUsed() / config.getBytesMovedFullRebalance());
-            moveAssignments.addAll(pushTasksOffHost(heavyHost, Collections.singletonList(host), true, byteLimitFactor,
-                                                    config.getTasksMovedFullRebalance(), true));
+            List<JobTaskMoveAssignment> assignments =
+                    getMoveAssignments(heavyHost, lightHostList, true, byteLimitFactor, moveLimit, true);
```
yuesong: If you use `1` as the `moveLimit` for `getMoveAssignments`, it would be clearer that you are only moving at most one task from each `heaveHost`. Either way, some comments would help others understand what the code does.

apsubra: Retained moveLimit with the original value to allow other moveAssignments to be added if the 1st one is rejected and added javadoc to explain what the method does.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
```diff
@@ -55,14 +53,16 @@ public boolean isEmpty() {
...
     public boolean add(JobTaskMoveAssignment assignment) {
-        List<JobKey> jobKeysToTargetHost = moveAssignmentList.stream()
-                          .filter(moveAssignment -> moveAssignment.getTargetUUID().equals(assignment.getTargetUUID()))
-                          .map(moveAssignment -> moveAssignment.getJobKey())
-                          .collect(Collectors.toList());
+        // Check if moveAssignmentList contains an assignment that moves a replica of the task to the same target host
+        boolean isJobKeyPresent = this.moveAssignmentList.stream()
+                                                         .anyMatch(moveAssignment ->
+                                                                           moveAssignment.getJobKey().equals(assignment.getJobKey()) &&
+                                                                           moveAssignment.getTargetUUID().equals(assignment.getTargetUUID()));
```
yuesong: Must add condition `moveAssignment.getTargetUUID() != null`. Also needs better indention :)

yuesong:
I'm good with the code now. :)

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
tedpearson: Add a comment on your for loop that specifically states that only one assignment will be added.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -1474,7 +1512,7 @@ private JobTaskItemByHostMap generateTaskCountByHost(List<HostState> hosts, Iter
         Map<String, Boolean> snapshot = new HashMap<>(recentlyBalancedHosts.asMap());
         for (JobTaskMoveAssignment assignment : candidateAssignments) {
             String newHostID = assignment.getTargetUUID();
-            if (newHostID == null) {
+            if (newHostID == null && !assignment.delete()) {
```
tedpearson: Can you explain what this change does?

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
```diff
@@ -29,12 +33,38 @@
...
+        return this.moveAssignmentList;
+    }
+
+    /**
+     * Add the given JobTaskMoveAssignment to the moveAssignmentList
+     * if the assignment does not move a replica of the same task to the same target host
+     * @param assignment JobTaskMoveAssignment to be added to the moveAssignmentList
+     * @return <tt>true</tt> if this JobTaskMoveAssignment was added to the moveAssignmentList
+     */
     public boolean add(JobTaskMoveAssignment assignment) {
```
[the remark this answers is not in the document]
apsubra: Current uses of MoveAssignmentList have a small ArrayList (of size 6 or 1). :)

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -419,31 +450,10 @@ public boolean canReceiveNewTasks(HostState host) {
...
-                }
-                if (rv.size() >= moveLimit) {
-                    break;
-                }
-            }
-        }
-        return rv;
+        List<JobTaskMoveAssignment> moveAssignments = getMoveAssignments(host, otherHosts, limitBytes, byteLimitFactor, moveLimit, obeyDontAutobalanceMe);
+        markRecentlyReplicatedTo(moveAssignments);
+        moveAssignments.addAll(purgeMisplacedTasks(host, moveLimit));
```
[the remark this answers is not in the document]
apsubra: purgeMisplacedTasks returns a list of assignments with assignment.targetUUID = null (for non-existent jobs). These were pruned as an assignments with no chance of happening; and orphaned tasks were not added as delete assignments. This change should allow such tasks to be marked for deletion.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
apsubra: Delete assignments (from purgeMisplacedTasks) have targetUUID = null and assignment.delete() = true. This change prevents pruning delete assignments.
@yuesong reverting this change to an earlier commit because: 
when newHostID == null, assignment.delete() == true. The assignment is pruned by the following check:
`if ((newHost == null) || newHost.hasLive(jobKey) || !canReceiveNewTasks(newHost)) {
                log.warn("[spawn.balancer] decided not to move task from job {} to host {} " +
                         "because it cannot receive the new task", jobID, newHostID);`

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/MoveAssignmentList.java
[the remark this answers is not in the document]
tedpearson: 👍

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancer.java
```diff
@@ -450,7 +449,8 @@ public boolean canReceiveNewTasks(HostState host) {
                                                          double byteLimitFactor,
                                                          int moveLimit,
                                                          boolean obeyDontAutobalanceMe) {
-        List<JobTaskMoveAssignment> moveAssignments = getMoveAssignments(host, otherHosts, limitBytes, byteLimitFactor, moveLimit, obeyDontAutobalanceMe);
+        long byteLimit = (long) byteLimitFactor * config.getBytesMovedFullRebalance();
```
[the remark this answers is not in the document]
apsubra: 👍

tedpearson:
All my concerns are fixed. :)

PR #266, opened 2017-08-11
Murmur3 hash will be using through thread safe singleton ref T69516

satang500:
using on-demand holder pattern

### DIFF
--- a/hydra-essentials/src/main/java/com/addthis/hydra/common/hash/Murmur3HashFunction.java
+++ b/hydra-essentials/src/main/java/com/addthis/hydra/common/hash/Murmur3HashFunction.java
@@ -15,7 +15,21 @@ package com.addthis.hydra.common.hash;
 
 import com.google.common.hash.Hashing;
 
-final class Murmur3HashFunction implements StringAndByteHashFunction {
+public final class Murmur3HashFunction implements StringAndByteHashFunction {
+
+    private Murmur3HashFunction() {}
+
+    private static class HolderClass {
+        private static final Murmur3HashFunction instance = new Murmur3HashFunction();
+    }
+
+    public static Murmur3HashFunction getInstance() {
+        return HolderClass.instance;
+    }
+
+    public static int getHash(String input) {
+        return Hashing.murmur3_32().hashUnencodedChars(input).asInt();
+    }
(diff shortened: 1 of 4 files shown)


hydra-essentials/src/main/java/com/addthis/hydra/common/hash/Murmur3HashFunction.java
```diff
@@ -15,7 +15,24 @@
...
+        return HolderClass.instance;
+    }
+
+    public static int getHash(String input) {
+        return Hashing.murmur3_32().hashUnencodedChars(input).asInt();
+    }
+
+    public static int getHash(byte[] input) {
+        return Hashing.murmur3_32().hashBytes(input).asInt();
+    }
```
yuesong: This can simply be:
```
public final class Murmur3HashFunction implements StringAndByteHashFunction {

    private static final Murmur3HashFunction INSTANCE = new Murmur3HashFunction();

    public static Murmur3HashFunction getInstance() {
        return INSTANCE;
    }
    ...
}
```

satang500: There are multiple ways for thread safe singleton. I prefer one of the ways which I used since it is known to be fast next to enum, but if you prefer shorter code by using static field initialization, I have no problem to modify it.

yuesong:
The two additional `getHash` methods aren't necessary.

As for the singleton pattern, I failed to see any benefit of using an extra `HolderClass` class. I've never seen any usage of that. I could be wrong though. Do you have a web page or reference you can share on that pattern? Always happy to learn something new.

satang500:
yeah, I forgot to remove unnecessary methods thanks for catching it.

There was a web page which summarizes different ways of thread safe singleton and pros and cons for each approach. But I lost lots of bookmarks recently. For on-demand holder, there is a wiki, https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom

yuesong:
Thanks for sharing the link. Lazy initialization is unnecessary here as the class is really lightweight.

Please update the javadoc for the new type 5. Otherwise the change looks good now.

PR #267, opened 2017-08-14
Alias

satang500:


hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -13,35 +13,83 @@
...
 public class AliasManagerImpl implements AliasManager {
-
     private static final Logger log = LoggerFactory.getLogger(AliasManagerImpl.class);
+    private Map<String, String> aliases;
+    /* This SpawnDataStore must be the same type (zookeeper/priam) between Spawn and Mqmaster. This should
+         * be guaranteed by the implementation of DataStoreUtil. */
+    public static final String ALIAS_PATH = "/query/alias";
+    private final SpawnDataStore spawnDataStore;
+    private final HashMap<String, List<String>> alias2jobs;
+    private final HashMap<String, String> job2alias;
```
jhorwit2: Why not use a [BiMap](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/BiMap.html)? That's cleaner IMO.

hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -13,35 +13,83 @@
...
+        this.ac = new AliasCache();
+    }
 
-    public AliasManagerImpl(SpawnDataStore spawnDataStore) {
-        this.aliasBiMap = new AliasBiMap(spawnDataStore);
-        aliasBiMap.loadCurrentValues();
+    public AliasManagerImpl(SpawnDataStore spawnDataStore) throws Exception {
+        this.spawnDataStore = spawnDataStore;
+        this.mapLock = new ReentrantLock();
+        this.mapper = new ObjectMapper();
```
jhorwit2: This should be declared statically.

hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -51,14 +99,158 @@ public AliasManagerImpl(SpawnDataStore spawnDataStore) {
...
         }
     }
 
+    public void putAlias(String alias, List<String> jobs) {
+        mapLock.lock();
+        try {
+            alias2jobs.put(alias, jobs);
+            job2alias.put(jobs.get(0), alias);
+            StringWriter sw = new StringWriter();
+            mapper.writeValue(sw, jobs);
```
jhorwit2: If you use `writeValueAsString(jobs)` you wont need a `StringWriter`

hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -51,14 +99,158 @@ public AliasManagerImpl(SpawnDataStore spawnDataStore) {
...
+    /**
+     * Refresh an alias based on datastore
+     *
+     * @param alias The alias to refresh
+     */
+    private void refreshAlias(String alias) {
+        try {
+            updateAlias(alias, this.spawnDataStore.getChild(ALIAS_PATH, alias));
+        } catch (ExecutionException e) {
+            log.warn("Failed to refresh alias: {}", alias, e);
```
jhorwit2: why isn't this an error?

hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -51,14 +99,158 @@ public AliasManagerImpl(SpawnDataStore spawnDataStore) {
...
+        } catch (Exception e) {
+            log.error("",e);
```
jhorwit2: "" isn't super useful for debugging. Something like `Unexpected error refreshing alias <alias>` would be better.

hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -51,14 +99,158 @@ public AliasManagerImpl(SpawnDataStore spawnDataStore) {
...
+        }
+    }
+
+    /**
+     * Load the jobIds for a particular alias from the SpawnDataStore
+     *
+     * @param alias The alias key to check
+     * @return String The data that was updated (so the cache can be updated)
+     */
+    @Nullable private String updateAlias(String alias, @Nullable String data) {
```
jhorwit2: This method has unclear side effects IMO. It seems to do an update or delete based on whether or not the data argument is passed. The java docs need to be updated.

hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -51,14 +99,158 @@ public AliasManagerImpl(SpawnDataStore spawnDataStore) {
...
+        if (Strings.isNullOrEmpty(alias)) {
+            return data;
+        }
+        if (Strings.isNullOrEmpty(data)) {
+            deleteAlias(alias);
+            return data;
+        }
+        List<String> jobs = decodeAliases(data);
+        if (jobs.isEmpty()) {
+            log.warn("no jobs for alias {}, ignoring {}", alias, alias);
```
jhorwit2: `ignoring alias {} since there are no jobs` or something would avoid the redundant alias log entry.

hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -51,14 +99,158 @@ public AliasManagerImpl(SpawnDataStore spawnDataStore) {
...
+        } finally {
+            mapLock.unlock();
+        }
+        return data;
+    }
+
+    @VisibleForTesting
+    protected List<String> decodeAliases(@Nonnull Object data) {
+        try {
+            return mapper.readValue(data.toString(), new TypeReference<List<String>>() {});
```
jhorwit2: Is data always a string here? If so why is it an object paramater?

hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -51,14 +99,158 @@ public AliasManagerImpl(SpawnDataStore spawnDataStore) {
...
+        }
     }
 
+    /**
+     * Test a job/alias pair to see if an alias has disappeared
+     *
+     * @param job   The job to test
+     * @param alias The alias to check
+     */
+    private void checkAlias(String job, String alias) {
```
jhorwit2: This method also has unclear side effects since it removes job from `job2alias` when the method name is called `checkAlias`

hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -51,14 +99,158 @@ public AliasManagerImpl(SpawnDataStore spawnDataStore) {
...
+        ac.deleteAlias(alias);
+    }
+
+    /**
+     * Get an alias for a particular jobId
+     *
+     * @param jobid The jobId to check
+     * @return One of the aliases for that job
+     */
+    public String getLikelyAlias(String jobid) {
```
jhorwit2: why likely alias? Wouldn't this be the alias?

hydra-main/src/main/java/com/addthis/hydra/job/alias/AliasManagerImpl.java
```diff
@@ -51,14 +99,158 @@ public AliasManagerImpl(SpawnDataStore spawnDataStore) {
...
+                for (String job : jobs) {
+                    String aliasVal = job2alias.get(job);
+                    if (Objects.equals(aliasVal, alias)) {
+                        job2alias.remove(job);
+                    }
+                }
+            }
+        } finally {
+            mapLock.unlock();
+        }
```
jhorwit2: Shouldn't you copy the maps first? What happens if the `deleteChild` call below fails? You now have a bad state.

hydra-main/src/main/java/com/addthis/hydra/query/spawndatastore/AliasCache.java
```diff
@@ -0,0 +1,104 @@
...
+            log.error("There is no job for alias {} ", alias);
+            return null;
+        }
+        return jobs;
+    }
+
+   public void deleteAlias(String alias) {
+        mapCache.remove(alias);
+    }
+}
```
satang500: FYI, There is weird symbol right next to right curly bract in line 104. Cannot remove it from PR, but it's ok in source code.

tedpearson: If you mouse over the 'weird symbol' it will tell you what it means. Please try this.

PR #268, opened 2017-09-19
[WIP] changed the file writing mode to avoid an empty minion.state wh…

satang500:
- Changed such that file can be appended except when minion starts (which is the same in the original code)
- added minion close before shutdown when TaskRunner run throws an exception

There are two places 
(run method in  CommandTaskDeleteRunner and kickNextJob method in Minion) 
where they use minionStateLock again 
even though writeState() is already locking via minionStateLock.
Since this is ReentrantLock, it's no harm, but wondering why they use this way.

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/minion/Minion.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/minion/Minion.java
@@ -278,7 +278,7 @@ public class Minion implements MessageListener<CoreMessage>, Codable, AutoClosea
             if (liveEverywhereMarkerFile.createNewFile()) {
                 log.info("cutover to live-everywhere tasks");
             }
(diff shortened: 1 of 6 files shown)


hydra-main/src/main/java/com/addthis/hydra/minion/Minion.java
```diff
@@ -581,10 +581,10 @@ private boolean updateTaskMeta(String jobID, File taskRoot) throws IOException {
         }
     }
 
-    void writeState() {
+    void writeState(boolean append) {
```
ranadessr: Is append the correct thing to do? Minion state is a JSON object. If you keep appending, wont that result in invalid json in the file? Have you tested that?

PR #269, opened 2017-09-29
Deploy script web asset

satang500:
Change:
If any file in web asset directory changed, 
   builds web asset on you local and then proceeds normal deployment.
else 
  proceeds normal deployment as we used to do

NOTE: 
- didn't arc diff and just pushed internal hydra,  so can't see PR in phabricator. 
   It has the same branch name as oss.
- Changes in app.js is for testing temporarily as I explained in scrum to see if it works on production as well.
- Because even though tested on test cluster, I'd like to see if it's working correctly in production as well.
- If it works production, then will remove it in the next deploy.


tedpearson:
Where can i find the changes to deploy.py?

satang500:
already mentioned in the first bullet in the NOTE in my comment above.

satang500:
static web assets in hydra-main/web/spawn2/  ( except hydra-main/web/spawn2/build ) can be also modified. But this static web asset will be extremely rarely modified, so that's why entire hydra-main/web/spawn2/ was ignored which is my guess. So for convenience I can revert .gitignore as before to avoid some untracked file/dir issue for your or for others (even though I don't have this issue though ).

PR #272, opened 2017-11-15
Ad aware 2

apsubra:


hydra-main/src/main/java/com/addthis/hydra/job/HostFailWorker.java
```diff
@@ -240,7 +241,8 @@ private void failNextHost() {
             if (failState == FailState.FAILING_FS_DEAD) {
                 // File system is dead. Relocate all tasks ASAP.
                 markHostDead(failedHostUuid);
-                spawn.getSpawnBalancer().fixTasksForFailedHost(spawn.hostManager.listHostStatus(null), failedHostUuid);
+                HostLocation location = spawn.hostManager.getHostState(failedHostUuid).getHostLocation();
+                spawn.getSpawnBalancer().fixTasksForFailedHost(spawn.hostManager.listHostStatusByZone(null, location), failedHostUuid);
```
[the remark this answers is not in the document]
apsubra: Does an ordering of hosts by zone sound better instead of listHostStatusByZone in this case?

hydra-main/src/main/java/com/addthis/hydra/minion/Zone.java
```diff
@@ -0,0 +1,76 @@
+package com.addthis.hydra.minion;
```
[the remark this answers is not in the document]
apsubra: It's not currently used in the project - will remove it :)

tedpearson:
Currently the code could allow rebalances to end up with all live tasks + replicas for a task on the same AD or physical host. We'll need to refactor the code to make sure this isn't possible. (specifically: SpawnBalancer.isAvailablilityDomainAware + SpawnBalancer.getMinReplicaZones)

tedpearson:
Let's also see how much work it would be to spread out tasks (not replicas) when a job is created, evenly across ADs.

tedpearson:
The changes to SpawnBalancer.getAssignmentsForNewReplicas are going to make all the replicas be spread across only a few hosts. The old logic of using scoreHeap (which is NOT host score... it's number of replicas per host) and adding +1 each time a replica is added is okay, either use something similar or come up with a new approach.

tedpearson:
Just so we don't forget: generateTaskCountHostScoreMap (original code) appears to incorrectly use the inverse of host.getDiskUsedPercent.

hydra-main/src/main/java/com/addthis/hydra/job/spawn/AvailabilityDomain.java
```diff
@@ -0,0 +1,27 @@
+/*
```
kxingit: feel like this deserves a better name, because AvailabilityDomain is a thing in OCI. Maybe ProrityLevel or SpreadPrioirty or GeoLevel... ?

hydra-main/src/main/java/com/addthis/hydra/minion/HostLocation.java
```diff
@@ -50,7 +55,62 @@ public String getPhysicalHost() {
...
+    /**
+     * Returns a score of distance between this and another HostLocation.
+     * Zero means maximum distance.
+     * This method is used to sort HostLocations by their distance score.
+     */
+    public int assignScoreByHostLocation(HostLocation o) {
+        if (this.getDataCenter().equals(o.getDataCenter())) {
+            if (!this.getRack().equals(o.getRack())) {
+                // Same dataCenter, different rack
+                return AvailabilityDomain.RACK.score;
```
kxingit: don't understand how this works.. how are the returning score determined?

apsubra: Just picked a convention that lets us easily sort based on "AD-distance" score :)

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostCandidateIterator.java
```diff
@@ -0,0 +1,129 @@
...
+                    continue;
+                }
+                HostLocation hostLocation = host.getHostLocation();
+                int score = 0;
+                // HostLocation further away from task and replica locations gets lower score
+                for (HostLocation location : locations) {
+                    score += hostLocation.assignScoreByHostLocation(location);
+                }
+                double avgScore = locations.isEmpty()? 0 : (double) score / (double) locations.size();
+                if (avgScore == (double) this.minScore) {
```
kxingit: also don't understand how this works

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostCandidateIterator.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Get new replica host for a rebalancing task
+     * note: this is not used at the moment, but saved for a later date when we rewrite spawnbalancer
+     */
+    public List<String> getNewReplicaHosts(JobTask task, @Nullable String excludeHost) {
+        return this.getNewReplicaHosts(1, task, excludeHost, true);
+    }
+
+}
```
kxingit: it looks like it can return an empty list, while a better approach would be it can always give a list of size `replicas`, so that it will not block things

hydra-main/src/test/java/com/addthis/hydra/job/spawn/balancer/SpawnBalancerTest.java
```diff
@@ -593,6 +763,107 @@ public  void moveAssignmentListTest() throws Exception {
...
+            assertTrue("Should choose HostLocation with min score and different datacenter",
+                       iterator.next().equals("hostId6"));
+            assertTrue("Should choose HostLocation on different datacenter next",
+                       iterator.next().equals("hostId2"));
+            assertTrue("Should choose Host with lower score on different rack next",
+                       iterator.next().equals("hostId5"));
+            assertTrue("Should choose Host on different physical host next",
+                       iterator.next().equals("hostId3"));
+            assertTrue("Should not choose Host in the same location if other hosts available",
+                       iterator.next().equals("hostId2"));
```
kxingit: i think we should write some check method, instead of checking like this

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostCandidateIterator.java
apsubra: An empty list is returned only when `replicas <= 0`; a list of size `replicas` is returned otherwise - am I missing some line of code where we return an empty list?

hydra-main/src/main/java/com/addthis/hydra/job/spawn/balancer/HostCandidateIterator.java
apsubra: @kxingit I've added more documentation for the scores and replica selection strategy used here - is there something you're concerned about with replica selection?

PR #273, opened 2017-12-08
Host location with di

kxingit:


hydra-main/src/main/java/com/addthis/hydra/minion/FileReaderHostLocationInitializer.java
```diff
@@ -0,0 +1,58 @@
...
+                dataCenter = config.getString("dataCenter");
+            }
+            if (config.hasPath("rack")) {
+                rack = config.getString("rack");
+            }
+            if (config.hasPath("physicalHost")) {
+                physicalHost = config.getString("physicalHost");
+            }
+        } catch (Exception e) {
+            log.warn("error getting host location from {}: {}", file.getName(), e);
```
tedpearson: You shouldn't use errors as a string interpolation in log methods, but you can pass it without interpolation and it should just do the right thing. Can explain in depth if needed.

hydra-main/src/main/java/com/addthis/hydra/minion/HostnameMatcherHostLocationInitializer.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @JsonCreator
+    public HostnameMatcherHostLocationInitializer() {
+        log.info("using HostnameMatcherHostLocationInitializer.");
+    }
+
+    public HostLocation getHostLocation() {
+
+        String hostname = Parameter.value("minion.localhost");
+
```
tedpearson: I'm not comfortable that the regex below are sufficient. Let's discuss how to make it better.

hydra-main/src/main/java/com/addthis/hydra/minion/Minion.java
```diff
@@ -246,6 +248,12 @@ private Minion(@JsonProperty("dataDir") File rootDir,
         } else {
             myHost = InetAddress.getLocalHost().getHostAddress();
         }
+        if (hostLocationInitializer == null) {
```
tedpearson: If this could be null, you should also make the parameter in the constructor `@Nullable`

hydra-main/src/main/java/com/addthis/hydra/minion/HostnameMatcherHostLocationInitializer.java
tedpearson: To recap our discussion,
- We should by default not have any regex, it should be specified in config
- Users should not be required to provide a regex for each availability property.
- The regexes should be more specific, for example for AD, something like this for OCI: `^[^\.]+\.(ad[\.]+).+`

PR #274, opened 2018-01-02
Add prometheus metrics for spawn, minion, mqmaster, and mqworker

kxingit:
Addressed comments in https://github.com/addthis/hydra/pull/270

### DIFF
--- a/hydra-main/pom.xml
+++ b/hydra-main/pom.xml
@@ -145,6 +145,23 @@
       <artifactId>metrics-ganglia</artifactId>
     </dependency>
 
+    <!-- prometheus metrics -->
+    <dependency>
+      <groupId>io.prometheus.jmx</groupId>
+      <artifactId>collector</artifactId>
+      <version>0.10</version>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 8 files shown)


tedpearson:
If we do the things that MetricsServletMaker was doing for metricsServlet, are all the handler.start/stop things still required?

kxingit:
I think the original MetricsServletMaker that is supposed to handle "metrics" never worked properly. So if I run master branch and then curl: 
$ curl localhost:5555/ping
ACK
$ curl localhost:5555/metrics/
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
<title>Error 503 Service Unavailable</title>

tedpearson:
Confirmed in production. Funky.

One more minor thing. Can we rename `handler` since it's not obvious from the name what it does.

tedpearson:
Does this still need a review @kxingit ?

kxingit:
@tedpearson nope, another branch has been merged https://github.com/addthis/hydra/pull/270
this branch can be deleted

PR #275, opened 2018-03-02
Upgrade jetty 8 to 9

yuesong:


hydra-main/src/main/java/com/addthis/hydra/job/web/SpawnService.java
```diff
@@ -39,93 +39,125 @@
...
-        spawn.getSystemManager().updateSslEnabled(sslEnabled);
+    }
+
+    private SslConnectionFactory createSslConnectionFactory(SpawnServiceConfiguration configuration)
+            throws IOException {
+        SslContextFactory sslContextFactory = new SslContextFactory();
+        sslContextFactory.setKeyStorePath(configuration.keyStorePath);
+        sslContextFactory.setKeyStorePassword(readFile(configuration.keyStorePassword));
+        sslContextFactory.setKeyManagerPassword(readFile(configuration.keyManagerPassword));
+        return new SslConnectionFactory(sslContextFactory, HttpVersion.HTTP_1_1.asString());
```
jhorwit2: Can you do http2?

yuesong: Jetty has http2 support, but I don't know much of it and what that entails for hydra

jhorwit2: It should be invisible for Hydra if you enable it.

yuesong: Turns out that enabling HTTP 2 on the server side is non-trivial amount of work, especially for TLS:

http://www.eclipse.org/jetty/documentation/9.4.8.v20171121/http2.html

Going to leave it as a potential future improvement for now.

PR #276, opened 2018-03-02
Improve readability of clusters-jobs section

wla80:
minor improvement of the text


wla80:
Hi @tedpearson thanks for the feedback. I just fixed my commit based on your comments. Thanks

PR #278, opened 2018-04-27
Rabbitmq dependency updates

Tingting-He-ODC:
3.4.1->5.2.0
Major change was the removal of QueueingConsumer.java. Hydra RabbitQueueingConsumer class now is implementing DefaultConsumer directly.


tedpearson:
I don't think I can review this until the dependency update gets merged.

hydra-mq/src/main/java/com/addthis/hydra/mq/RabbitQueueingConsumer.java
```diff
@@ -14,14 +14,43 @@
...
     public RabbitQueueingConsumer(Channel channel) {
         super(channel);
+        _queue = new LinkedBlockingQueue<Delivery>();
+    }
+
+    @Override public void handleDelivery(String consumerTag,
+            Envelope envelope,
+            AMQP.BasicProperties properties,
+            byte[] body)
+            throws IOException
```
tedpearson: IOException is never thrown

hydra-mq/src/main/java/com/addthis/hydra/mq/RabbitQueueingConsumer.java
```diff
@@ -14,14 +14,43 @@
...
+import java.util.concurrent.LinkedBlockingQueue;
 
-public class RabbitQueueingConsumer extends QueueingConsumer {
+public class RabbitQueueingConsumer extends DefaultConsumer {
 
+    private final BlockingQueue<Delivery> _queue;
 
     public RabbitQueueingConsumer(Channel channel) {
         super(channel);
+        _queue = new LinkedBlockingQueue<Delivery>();
```
tedpearson: can use diamond `<>`

hydra-mq/src/main/java/com/addthis/hydra/mq/RabbitQueueingConsumer.java
```diff
@@ -14,14 +14,43 @@
...
+import com.rabbitmq.client.Envelope;
+import com.rabbitmq.client.AMQP;
+import java.io.IOException;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
 
-public class RabbitQueueingConsumer extends QueueingConsumer {
+public class RabbitQueueingConsumer extends DefaultConsumer {
 
+    private final BlockingQueue<Delivery> _queue;
```
tedpearson: Why the underscore? I would prefer not to use underscores in variable names.

Tingting-He-ODC:
The old QueueingConsumer:

`public class QueueingConsumer extends DefaultConsumer {
    private final BlockingQueue<Delivery> _queue;

    // When this is non-null the queue is in shutdown mode and nextDelivery should
    // throw a shutdown signal exception.
    private volatile ShutdownSignalException _shutdown;
    private volatile ConsumerCancelledException _cancelled;

    // Marker object used to signal the queue is in shutdown mode.
    // It is only there to wake up consumers. The canonical representation
    // of shutting down is the presence of _shutdown.
    // Invariant: This is never on _queue unless _shutdown != null.
    private static final Delivery POISON = new Delivery(null, null, null);

    public QueueingConsumer(Channel ch) {
        this(ch, new LinkedBlockingQueue<Delivery>());
    }

    public QueueingConsumer(Channel ch, BlockingQueue<Delivery> q) {
        super(ch);
        this._queue = q;
    }

    @Override public void handleShutdownSignal(String consumerTag,
                                               ShutdownSignalException sig) {
        _shutdown = sig;
        _queue.add(POISON);
    }

    @Override public void handleCancel(String consumerTag) throws IOException {
        _cancelled = new ConsumerCancelledException();
        _queue.add(POISON);
    }

    @Override public void handleDelivery(String consumerTag,
                               Envelope envelope,
                               AMQP.BasicProperties properties,
                               byte[] body)
        throws IOException
    {
        checkShutdown();
        this._queue.add(new Delivery(envelope, properties, body));
    }
... ...
... ...
}
`

PR #281, opened 2018-06-11
Constrain minion log bytes to about 20MB.

Tingting-He-ODC:
Limiting log byte[] to size 20M. Print out warning if exceeds.

### DIFF
--- a/hydra-main/src/main/java/com/addthis/hydra/minion/LogUtils.java
+++ b/hydra-main/src/main/java/com/addthis/hydra/minion/LogUtils.java
@@ -35,7 +35,7 @@ import org.slf4j.LoggerFactory;
 
 public final class LogUtils {
     private static final Logger log = LoggerFactory.getLogger(LogUtils.class);
-    private static final int logBufLimit = 20000000;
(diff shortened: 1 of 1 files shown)


hydra-main/src/main/java/com/addthis/hydra/minion/LogUtils.java
```diff
@@ -93,8 +94,16 @@ public static JSONObject readLogLines(File file, int startOffset, int lines) {
                     }
                 }
                 bytesRead = (int) (len - off);
-                byte[] buf = new byte[bytesRead];
-                raf.read(buf);
+                byte[] buf = null;
+                // limiting log reads below 20MB, in case of reaching heap limit and crashing minion
+                if(bytesRead >= logBufLimit) {
```
tedpearson: Each of these 4 sections could be replaced by a call to a new method you can create. the raf.seek() calls can be made before the call to the method. Something like "readFileBytesLimited(long requestedBytes)" which returns a new byte[] created in the method.

tedpearson:
Looks good. Let's wait to merge until after the next successful hydra release.

PR #282, opened 2018-06-14
Rolling restart query system

apsubra:


hydra-main/src/main/java/com/addthis/hydra/query/web/HttpQueryHandler.java
```diff
@@ -284,6 +284,32 @@ private void fastHandle(ChannelHandlerContext ctx, FullHttpRequest request, Stri
...
+                } else {
+                    writer.write("[0]");
+                }
+                break;
+            }
+            case "/mqmaster/healthcheck": {
+                if(meshQueryMaster.getIsActive().get()) {
+                    writer.write("[1]");
+                } else {
+                    writer.write("[0]");
```
tedpearson: I'm curious as to why the brackets around the numbers in the output here?

hydra-main/src/main/java/com/addthis/hydra/query/MeshQueryMaster.java
```diff
@@ -126,6 +130,18 @@ public MeshQueryMaster(QueryTracker tracker) throws Exception {
...
+    public AtomicBoolean getIsActive() {
+        return isActive;
+    }
+
+    public boolean deactivateMqMaster(boolean deactivate) {
+        if(deactivate) {
+            // Set mqMaster's isActive field to false
+            return this.isActive.compareAndSet(true, !deactivate);
+        }
+        return this.isActive.compareAndSet(false, true);
```
tedpearson: This whole method could be replaced with `this.isActive.compareAndSet(deactivate, !deactivate)`. What are we doing on the client side if compare and set is not successful?

hydra-main/src/main/java/com/addthis/hydra/query/web/HttpQueryHandler.java
```diff
@@ -284,6 +284,32 @@ private void fastHandle(ChannelHandlerContext ctx, FullHttpRequest request, Stri
...
+                        break;
+                    default:  success = false;
+                }
+                if(success) {
+                    writer.write("[1]");
+                } else {
+                    writer.write("[0]");
+                }
+                break;
+            }
```
tedpearson: We should just set deactivate to true or false, and not worry about sending back success or state. If we get an http 200, assume all is well.

hydra-main/src/main/java/com/addthis/hydra/query/web/HttpQueryHandler.java
apsubra: The plan was to have the mq masters under a load balancer. The lb we received (at the time, Ops created it for us) was configured to use "[1] or [0]" as the healthcheck.

apsubra: Edited healthcheck to be Http 200/400 for now, I made a note to look at this once again when there's more clarity on what the lb will look like

hydra-main/src/main/java/com/addthis/hydra/query/web/HttpQueryHandler.java
apsubra: Edited deactivate endpoint to just set() and return Http ok/400

hydra-main/src/main/java/com/addthis/hydra/query/web/HttpQueryHandler.java
```diff
@@ -284,6 +284,35 @@ private void fastHandle(ChannelHandlerContext ctx, FullHttpRequest request, Stri
                 writer.write(q.getPaths()[0]);
                 break;
             }
+            case "/mqmaster/deactivate": {
+                log.info("Received MeshQueryMaster deactivate request");
+                String shouldDeactivate = kv.getValue("deactivate");
+                switch (shouldDeactivate) {
```
tedpearson: Could throw NPE here.

hydra-main/src/main/java/com/addthis/hydra/query/web/HttpQueryHandler.java
```diff
@@ -284,6 +284,35 @@ private void fastHandle(ChannelHandlerContext ctx, FullHttpRequest request, Stri
...
+                                if (ctx.channel().isActive()) {
+                                    sendError(ctx, HttpResponseStatus.BAD_REQUEST);
+                                }
+                                return;
+                }
+                break;
+            }
+            case "/mqmaster/healthcheck": {
+                // Send 400 if mqMaster is not active
+                // Send 200 if mqMaster is active (NO-OP here as response OK is sent below)
```
tedpearson: Good comment!

hydra-main/src/main/java/com/addthis/hydra/query/web/HttpQueryHandler.java
apsubra: Added a null-check

tedpearson:
Looks good

PR #285, opened 2018-08-15
Added spawn.datastore.type: MYSQL2 to use mysql connectorJ.

Tingting-He-ODC:
Adding mysql connectorJ (in addition to current drizzle driver) option to be compatible with oci mysql version.


Tingting-He-ODC:
Have to use a mysql driver that is compatible with oci mysql8.0. Current drizzle driver does not have newer versions supporting mysql8.0, therefore switched to connectorJ. Have not tried compatibility with non-oci mysql version, since drizzle works perfectly with older mysql for now.

PR #286, opened 2018-09-11
Log4j config to let hydra log to file

jiayangli2:


hydra-uber/src/main/resources/log4j2.conf
```diff
@@ -10,26 +8,31 @@ _status_min = ${?logging.status}
...
     },
-    { name: STDERR,
-      type: Console,
+    { name: File,
+      type: RollingRandomAccessFile,
+      fileName: "${sys:hydra.logdir:-log}/${sys:logfile.name}.log",
+      filePattern: "${sys:hydra.logdir:-log}/old/${sys:logfile.name}-%d{yyMMddHHmmss}.log.gz",
       PatternLayout: { pattern: "%p [%t] %d{ISO8601} %F (line %L) %.-5000m%n" },
-      target : SYSTEM_ERR
+      OnStartupTriggeringPolicy: { minSize: 1 }
```
tedpearson: Let's discuss if we should use daily rolling instead of every startup.

hydra-uber/src/main/resources/log4j2.conf
```diff
@@ -41,7 +44,7 @@ loggers: {
     { name: org.apache.xbean.spring, level: warn }
   ],
   root: { level: ${_out_min}, AppenderRef: [
-    { ref: STDERR, level: ${_err_min}},
-    { ref: STDOUT, level: ${_out_min}}
+    { ref: File,   level: ${_err_min}},
+    { ref: Async,  level: ${_out_min}}
```
tedpearson: I think we should use a single appender, I think this current config will double-log error level messages.

hydra-uber/src/main/resources/log4j2.conf
jiayangODC: Decided to use daily rolling and set the period to 86400 seconds which is 24 hours. Log4j will find the smallest unit in filePattern and use that unit as the unit for the parameter in the TimeTriggeringPolicy. Since the smallest unit I used in filePattern is ss (seconds) that means I need to put 86400 to represent 1 day.

hydra-uber/src/main/resources/log4j2.conf
jiayangODC: I agree that the current config would potentially double-log the error level messages. However, after examining the log file generated on a test cluster I found the [WARN] logs were logged just once.

I suspect the hocon property ${?logging.stderr} was set to "off" somewhere in the codebase, which means _err_min would be "off" instead of "warn". Thus, the err msg would just go through the "info" channel which is the async channel.

 I also think that the old config would potentially log the error level twice but because of the same reason (_err_min = ${?logging.stderr} = "off"), the log it generated actually had just one copy of the [WARN] msg. If you examine https://github.com/addthis/hydra/blob/master/hydra-uber/src/test/java/com/addthis/hydra/uber/LogTest.java you and run the test in Intellij you will find that the [WARN] is printed out twice and I believe that should be the case if ${?logging.stderr} is not set to "off"

jiayangODC: The decision I made was to keep change { ref: File,   level: ${_err_min}} into { ref: File,   level: ${_out_min}} and delete the async appenderRef

PR #289, opened 2018-10-24
WIP: Jiayang/gdpr

jiayangODC:


tedpearson:
Some recommendations here:
- You're creating a 5-thread pool for something that will only happen one-at-a-time.
- I think a scheduled executor with fixed delay would be better suited to this use case, instead of using Thread.sleep to enforce the wait period. MoreExecutors even has a method that returns an exiting one.
- It might make sense to name the `refreshInterval` something that makes it obvious what the time units are, such as `refreshIntervalMin` or something.

hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/AbstractMatchStringFilter.java
```diff
@@ -258,6 +268,19 @@ public boolean passedFind(String sv) {
...
     }
 
+    @Override public void postDecode() {
+        if (refreshMinutes == 0) {
+            postDecodeHelper();
+        } else {
+            ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(1);
+            // upon JVM termination, wait for the tasks for up to 100ms before exiting the executor service
+            ScheduledExecutorService executorService = MoreExecutors.getExitingScheduledExecutorService(executor, 100, TimeUnit.MILLISECONDS);
+            executorService.scheduleWithFixedDelay(() -> {
```
tedpearson: My intellij inspections recommend using a method reference instead of a lambda here: `this::postDecodeHelper`

hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/AbstractMatchStringFilter.java
```diff
@@ -258,6 +268,19 @@ public boolean passedFind(String sv) {
         }
     }
 
+    @Override public void postDecode() {
+        if (refreshMinutes == 0) {
+            postDecodeHelper();
+        } else {
```
tedpearson: I noticed a pretty serious issue here - in the case where refresh is enabled, we only call postDecodeHelper in the scheduled executor service. While it's true that we call it immediately, it's called asynchronously without blocking. We need to block here so that this value filter can't be used until the post decode step has occured at least once. Therefore, we should always call postDecodeHelper in this method, and change the scheduled executor to not refresh immediately.

hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/AbstractMatchStringFilter.java
jiayangODC: Done

hydra-filters/src/main/java/com/addthis/hydra/data/filter/value/AbstractMatchStringFilter.java
jiayangODC: Done

PR #294, opened 2019-04-29
Mesh peer retry and vault changes.

Tingting-He-ODC:
Ignore the branch name... Mainly adding mesh peer retry and getting mysql password from vault.


Tingting-He-ODC:
Still need to change mesh release version first.

hydra-main/src/main/java/com/addthis/hydra/job/store/DataStoreUtil.java
```diff
@@ -109,6 +110,11 @@ public static SpawnDataStore makeSpawnDataStore(DataStoreType type) throws Excep
         Properties properties = new Properties();
         if (sqlUser != null) {
             properties.put("user", sqlUser);
+            if(sqlPasspath.length() != 0) {
```
tedpearson: use !isEmpty() :)

hydra-main/src/main/java/com/addthis/hydra/job/store/DataStoreUtil.java
```diff
@@ -109,6 +110,11 @@ public static SpawnDataStore makeSpawnDataStore(DataStoreType type) throws Excep
...
+                VaultWrapper vaultClient = (VaultWrapper) Class.forName("com.addthis.hydra.vault.VaultWrapperImpl").newInstance();//new VaultWrapperImpl();
```
tedpearson: Hard coding a classname that only exists in closed source is not the right way to do this.

hydra-main/src/main/java/com/addthis/hydra/job/store/DataStoreUtil.java
```diff
@@ -109,6 +110,11 @@ public static SpawnDataStore makeSpawnDataStore(DataStoreType type) throws Excep
...
+                vaultClient.login("https://vault.prd.phxshared.oracledatacloud.com");
```
tedpearson: Url should not be hard coded.

tedpearson:
Looks better

